PR #1, opened 2023-03-11
Contributing Guidelines

dlesnoff:
I wrote a simple CONTRIBUTE.md from the Python project and adapted it to Nim.

I advise for the use of snake_case for file_names, while I recommend camelCase and PascalCase for objects to follow [NEP1 specification](https://nim-lang.org/docs/nep1.html) of the language.

Third party modules is an empty file, but I add it in case, we need third party modules in the future.

I removed references to Gitter and check that the Discord link is correct in each file.

I added a MIT License similar to any of TheAlgorithms project with the correct year.

The README.md is minimalistic. If you have any enhancement that I can do, it might be better to take care of this in another PR, just so we show that the repository is already open to contributions.

- [ ] a better document structure
- [x] the use of Option
- [x] a relaxation on variable naming. We need to define use cases.
- [x] Effect system: use of `funcs` and raises pragma
- [x] Use of `unittest` and `runnableExamples`
- [x] Precise the necessity for a header with a correct file name
- [x] Specify that file names should be looked up in other repositories of The Algorithms
- [x] Specify compiler version (stable?, 2.0 rc1? -> Dev should add at least in PR description, the compiler version. In code too?)
- [x] Export marker. Mandatory?
I think it also lacks links to Nim resources like Nim 2.0 documentation.

PR #2, opened 2023-03-12
Implement "Check Anagram" algorithm

ZoomRmc:
Works only for us-ascii characters.
Builds and uses a counting associative array for comparison in one pass.
Tests and docs included.

Tried my best to follow the proposed guidelines as closely as possible.

### DIFF
--- a/strings/check_anagram.nim
+++ b/strings/check_anagram.nim
@@ -15,18 +15,14 @@
 ## Note: Generate full doc with `nim doc --docinternal check_anagram.nim`
 
 runnableExamples:
+
   doAssert "coder".isAnagram("credo")
 
   doAssert not "Nim".isAnagram("Nim")
 
   doAssert not "parrot".isAnagram("rapport")
 
-  let raisedException = try:
-      discard "Pearl Jam".isAnagram("Maple Jar")
-      false
-    except ValueError:
-      true
-  doassert raisedException
+  doAssertRaises(ValueError): discard "Pearl Jam".isAnagram("Maple Jar")
 
 ## Tests
 ## -----
@@ -41,9 +37,11 @@ runnableExamples:
 ##   is beneficial or detrimental to the performance, compared to char-by-char
 ##   approach of this module.
 
+{.push raises: [].}
+
 type
   Map = array[range['a'..'z'], Natural] ## An associative array with a direct
-                                        ## mapping from Char to a counter slot
+                                        ## mapping from Char to a counter slot.
 
 const
   UpperAlpha = {'A'..'Z'}
@@ -69,15 +67,15 @@ template normalizeChar(c: char) =
   elif c in UpperAlpha: toLowerUnchecked(c)
   else: raise newException(ValueError, "Character '" & c & "' is not a letter!")
 
-func isAnagram*(a, b: openArray[char]): bool {.raises: ValueError.} =
+func isAnagram*(wordA, wordB: openArray[char]): bool {.raises: ValueError.} =
   ## Checks if two words are anagrams of one another.
   ##
   ## Raises a `ValueError` on any non-letter character.
-  if a.len != b.len: return false
+  if wordA.len != wordB.len: return false
   var seenDifferent = false
   var mapA, mapB: Map
-  for chIdx in 0..<a.len:
-    let (chA, chB) = (normalizeChar(a[chIdx]), normalizeChar(b[chIdx]))
+  for chIdx in 0..<wordA.len:
+    let (chA, chB) = (normalizeChar(wordA[chIdx]), normalizeChar(wordB[chIdx]))
     # Identical characters xor to 0, must meet at least one 1 (difference)
     seenDifferent = bool(ord(seenDifferent) or (ord(chA) xor ord(chB)))
     mapA[chA].inc()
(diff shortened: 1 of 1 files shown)


strings/check_anagram.nim
```diff
@@ -0,0 +1,173 @@
...
+  doAssert not "Nim".isAnagram("Nim")
+
+  doAssert not "parrot".isAnagram("rapport")
+
+  let raisedException = try:
+      discard "Pearl Jam".isAnagram("Maple Jar")
+      false
+    except ValueError:
+      true
+  doassert raisedException
```
dlesnoff, line 29: I think top-level runnableExamples on top are great.
Proc-level RE are hard to distinguish from code but are pretty useful in the documentation of a large module.
Since most the Algorithms files are "snippets", I think we should not enforce proc-level RE, but enforce both a RE (for usage showcase) and a testsuite (for validity).
I will update the README.md accordingly.

dlesnoff:
I am fine with only US-Ascii characters on my side. We may handle Unicode algorithms in another file.

strings/check_anagram.nim
```diff
@@ -0,0 +1,173 @@
...
+type
+  Map = array[range['a'..'z'], Natural] ## An associative array with a direct
+                                        ## mapping from Char to a counter slot
+
+const
+  UpperAlpha = {'A'..'Z'}
+  LowerAlpha = {'a'..'z'}
+
+func toLowerUnchecked(c: char): char =
+  ## Lowers the case of an ASCII uppercase letter ('A'..'Z').
```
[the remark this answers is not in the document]
ZoomRmc: This is strictly an implementation detail so usually it's supposed to be a regular comment.

strings/check_anagram.nim
```diff
@@ -0,0 +1,173 @@
...
+  ## No checks performed, the caller must ensure the input is a valid character.
+  ##
+  ## See also:
+  ## * strutils.toLowerAscii
```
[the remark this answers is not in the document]
ZoomRmc: Done

strings/check_anagram.nim
```diff
@@ -0,0 +1,179 @@
...
+## Note: Generate full doc with `nim doc --docinternal check_anagram.nim`
+
+runnableExamples:
+  doAssert "coder".isAnagram("credo")
+
+  doAssert not "Nim".isAnagram("Nim")
+
+  doAssert not "parrot".isAnagram("rapport")
+
+  let raisedException = try:
```
pietroppeter, line 24: here you could use `doAssertRaises(ValueError)`, see https://nim-lang.org/docs/assertions.html#doAssertRaises.t%2Ctypedesc%2Cuntyped

strings/check_anagram.nim
```diff
@@ -0,0 +1,179 @@
+## Check Anagram
```
pietroppeter, line 1: depending or not whether [this suggestion](https://github.com/TheAlgorithms/Nim/pull/1#discussion_r1133207638) is accepted in contribution guidelines you could add a `{.push raises: [].}` at the top of the file. from a quick look the file should still compile (the assert raises - in debug mode - a Defect that is not considered an exception).

pietroppeter (approved):
a very nice contribution also very useful for testing the contribution guidelines!

strings/check_anagram.nim
```diff
@@ -0,0 +1,179 @@
...
+  else: raise newException(ValueError, "Character '" & c & "' is not a letter!")
+
+func isAnagram*(a, b: openArray[char]): bool {.raises: ValueError.} =
+  ## Checks if two words are anagrams of one another.
+  ##
+  ## Raises a `ValueError` on any non-letter character.
+  if a.len != b.len: return false
+  var seenDifferent = false
+  var mapA, mapB: Map
+  for chIdx in 0..<a.len:
```
dlesnoff, line 79: a, b single variable names :(
What are they ? Words. Why not wordA, wordB instead?
Apart from this style-cliving discussion, it seems to me that if they are openArrays, they may be indexed by something else: prefer a.low and a.high.
What if a and b are indexed differently ? Yes, you may have to do two additional sums when accessing b: b[low(b)-low(a)+chIdx]. We may suppose here that low(a) == low(b).

strings/check_anagram.nim
```diff
@@ -0,0 +1,179 @@
...
+func filterAnagrams*(word: string; candidates: openArray[string]): seq[string] {.raises: ValueError.} =
+  ## Checks if any of the candidates is an anagram of a `word` and returns them.
+  ##
+  ## Raises a `ValueError` on any non-letter character.
+  var wordMap: Map
+  var wordNormalized = newString(word.len)
+  for chIdx, c in word.pairs():
+    let c = normalizeChar(c)
+    wordNormalized[chIdx] = c
+    wordMap[c].inc()
```
dlesnoff, line 114: Ok I might relax a little bit the rule on single-character name for string algorithms only.
`character` is a pain to write. It shortens to `char` but it is a keyword (a pre-defined type). `charac` is very long and this variable is used a lot in string algorithms.

strings/check_anagram.nim
ZoomRmc: Here I'm following the Standard Library convention of naming arguments.

I'm not really sure what are you up to regarding indexing:

```nim
let a: array[range[10..13], char] = ['a', 'b', 'c', 'd']
let b: array[range[0..3], char] = ['w', 'x', 'y', 'z']

proc foo(a, b: openArray[char]) =
  for i in 0..3: echo a[i], b[i]

foo(a, b)
```

strings/check_anagram.nim
ZoomRmc: Thanks, missed it.

strings/check_anagram.nim
ZoomRmc: Well, there's just 5 entities in this block. Young kids, sleepy people and even I can keep 5 things in their head, at least for 4 lines. :D

I'd say, conventional `c` for char, `n` for number and `i` are acceptable when there's no competing variables in current scope.

strings/check_anagram.nim
dlesnoff: I thought that array indexes were passed along with the openArray into functions, sorry about this then.
I see nothing in NEP1 about single variable names, but I am nitpicking here.

dlesnoff (approved):
Thanks for this very good contribution !

strings/check_anagram.nim
ZoomRmc: I kind of understand the logic why the standard library uses `a`, `b`... for naming equivalent entities (where there's no clear relationship between them, unlike `src`-`dst` or `s`-`sub`), like in various `cmp` procs:
* It conveys these are the main subjects of the proc.
* It shows we don't know anything about the values beside its type.
* It conveys it's the raw input, not parsed/validated and frees the basic names like `word` to be used after that.
* It's generic so it works for this pattern regardless of the type.

When I start thinking about it, from the logic above follows in this case it's better to be explicit we expect not just any string. So I've talked myself into changing the names. :P

ZoomRmc: Well, my main reasoning is the func may fail on invalid input, so arg naming should reflect the kind of data expected: "Do not put any old string, it has to be a word!".

> Here I understand that the input has been stripped down of its extra characters.

I mean, we can shadow args in Nim of course, but even if we couldn't: `let month = a.parseMonth().toLower()` `a` is just some data which can be a `month`, means the proc has to deal with it in some way. Of course, it's all moot when you remember you should use types for this, but the complexity would quickly grow in case of a bunch of interleaved procs working on raw data (one example being `strutils`). So the correct approach always depends on the context of code usage.

PR #4, opened 2023-03-14
Create and populate the Maths directory

dlesnoff:
Add:
- maths/ directory
- aliquot_sum.nim
- addition_without_arithmetic.nim
- abs.nim
- ~~allocation_number.nim~~

These are examples adapted into Nim from [TheAlgorithms/Python/maths](https://github.com/TheAlgorithms/Python/tree/master/maths).

- [x] I will rebase from master to remove Contributing Guidelines and all these other stuffs when my other PR will be merged.
I tried to respect:
- {.push raises: [].} pragma on top of the file
- use of Natural(s) to limit exceptions -> might be improved by Positive type?
- use of Slices instead of Python tuples for ranges
- runnableExamples on top to showcase the functions of the module
- unittest under a when isMainModule block at the end

I respected also these but these are my personal conventions:
- ~~I do not export by default my functions.~~
- ~~I do not put a newline at the end of my files~~


ZoomRmc (changes requested):
Looks splendid!

maths/abs.nim
```diff
@@ -0,0 +1,87 @@
...
+  doAssert absMaxSort(@[3, -10, -2]) == -10
+
+func absVal(num: float): float =
+  ## Returns the absolute value of a number.
+  return if num < 0.0: -num else: num
+
+# Same for integers but return a Natural
+func absVal(num: int): Natural = (if num < 0: -num else: num)
+
+func absMin(x: seq[int]): Natural {.raises: [ValueError].} =
```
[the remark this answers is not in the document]
dlesnoff: I do not like the copy and the allocation but looks like callee's responsibilty for me.
At first, I just did want to translate the original in Python: https://github.com/TheAlgorithms/Python/blob/521fbca61c6bdb84746564eb58c2ef2131260187/maths/abs.py#L58
They sorted the whole list, without any question.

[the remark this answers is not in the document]
ZoomRmc: I don't think Python programs concern themselves with these things usually. :)

The main difference is if we require `seq` we force the caller to write boilerplate `toSeqs`/`@` when they have a view-slice (openarray) or just a regular array on hand.

Conceptually, here we return one number of the sequence, so we copy it all only to exploit the side-effect of sorting, for which we need the mutable state. Knowing that is probably of no concern of the caller.

This got me thinking, why copy and allocate at all?

```nim
func absMaxSort(x: openArray[int]): int {.raises: [ValueError].} =
  ## Returns the signed element whose absolute value is the largest in a sequence.
  var x: seq[int] = @x
  if x.len == 0:
    raise newException(ValueError, "Cannot find absolute maximum of an empty sequence")
  sort(x, proc (a, b: int): int = int(absVal(b)) - int(absVal(a)))
  return x[0]

func absMaxSortB(x: openArray[int]): int {.raises: [ValueError].} =
  ## Returns the signed element whose absolute value is the largest in a sequence.
  if x.len == 0:
    raise newException(ValueError, "Cannot find absolute maximum of an empty sequence")
  var (max, maxAbs) = (x[0], absVal(x[0]))
  for n in x:
    let nAbs = absVal(n)
    (max, maxAbs) = if nAbs > maxAbs: (n, nAbs) else: (max, maxAbs)
  max

import std/[strformat, stats, times, random]
template bench(name: string; samples: Positive; body: untyped) {.dirty.}=
  block:
    var stat: RunningStat
    var accum: uint # Prevents compiler from optimizing stuff away
    let globStart = cpuTime()
    for _ in 0 ..< samples:
      let start = cpuTime()
      accum.inc(body)
      stat.push cpuTime() - start
    echo name, ":"
    echo &"  Collected {stat.n} samples in {(cpuTime()-globStart):>4.3f} seconds"
    echo &"  Avg: {stat.mean * 1000.0 :>4.3f} ms  ",
      &"StDev: {stat.standardDeviationS * 1000 :>4.3f} ms  ",
      &"Min: {stat.min * 1000 :>4.3f} ms  ",
      &"Max: {stat.max * 1000 :>4.3f} ms"
    echo &"  Dummy sum of the returns to ensure `body` is not optimized away {accum}\n"

randomize()
var arr: array[32768, int]
for i in 0..<32768: arr[i] = rand(high(int))
bench("absMaxSort", 1000): arr.absMaxSort()
bench("absMaxSortB", 1000): arr.absMaxSortB()
```
```
absMaxSort:
  Collected 1000 samples in 5.267 seconds
  Avg: 5.263 ms  StDev: 4.516 ms  Min: 0.000 ms  Max: 21.000 ms
  Dummy sum of the returns to ensure `body` is not optimized away 18405476978678043616

absMaxSortB:
  Collected 1000 samples in 0.372 seconds
  Avg: 0.370 ms  StDev: 1.700 ms  Min: 0.000 ms  Max: 16.000 ms
  Dummy sum of the returns to ensure `body` is not optimized away 18405476978678043616
```

maths/addition_without_arithmetic.nim
```diff
@@ -0,0 +1,36 @@
+## Illustrate how to add the integer without arithmetic operation
```
pietroppeter, line 1:
```suggestion
## Illustrate how to implement addition of integers using bitwise operations
## See https://en.wikipedia.org/wiki/Bitwise_operation#Applications
```
I know this comes from the python version but the wording there to me is a bit weird (addition is an arithmetic operation, we are implementing here with bitwise operators instead of relying of the CPU instruction, which btw might be itself implemented with bitwise operators...). Also would add link to wikipedia which I think is relevant.

maths/aliquot_sum.nim
```diff
@@ -0,0 +1,25 @@
+{.push raises: [].}
```
pietroppeter, line 1: maybe add documentation comments with name of algorithm, a brief description (the first sentence from wikipedia are fine) and a link to wikipedia article https://en.wikipedia.org/wiki/Aliquot_sum

maths/allocation_number.nim
```diff
@@ -0,0 +1,52 @@
+## In a multi-threaded download, this algorithmm could be used to provide
```
pietroppeter, line 1: I could not find any reference to any "named algorithm", the only thing I see is indeed the algorithm in python (which has a different api since it returns a string with the range): https://github.com/TheAlgorithms/Python/blob/master/maths/allocation_number.py

honestly I would probably skip this, but in case you want to keep it:
- you might want to add a test with where numPartitions does not divide exactly numBytes
- you would need to pick a name of algorithm to have as first line in a documentation comment if we plan to merge this: https://github.com/TheAlgorithms/Nim/pull/5

pietroppeter (changes requested):
this is good thanks! my guess is you picked the first few files in alphabetical order from python repo? ;) there are some uncommon algorithms there (and my suggestion would be to skip allocation_number). :)

maths/addition_without_arithmetic.nim
dlesnoff: Many thanks for the link. The wording seems much more relevant this way!

maths/abs.nim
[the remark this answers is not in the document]
ZoomRmc: ping @dlesnoff 
Have you seen this? Any thoughts?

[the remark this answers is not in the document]
dlesnoff: Yes, sorry. I agree that the Python implementation was pretty bad from the start.
I think the point was to showcase how to sort the list with the named cmp parameter and an anonymous function.
I will use your absMaxSortB function instead.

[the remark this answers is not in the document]
dlesnoff: I added your function and its corresponding minimum version with their tests in the latest commit.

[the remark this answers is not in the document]
ZoomRmc: Yeah, walking the fine line between staying within the educational constraints and being idiomatic to the language you use is hard!

In the real world this is just a variation of a basic fold. Using an anonymous function in situations like this is usually fine, but here you need to store the absolute value of the accumulator and compare to that to avoid calling `accumulator.abs` on each comparison, so it becomes kind of pointless and longer than just a fold.

```nim
import sequtils, math;

func signedMinAbs(nums: openArray[int]): int =
  nums.mapIt((it, it.abs())).foldl(if a[1] < b[1]: a else: b)[0]

func signedMaxAbs(nums: openArray[int]): int =
  nums.mapIt((it, it.abs())).foldl(if a[1] > b[1]: a else: b)[0]

let nums = [3, -2, -1, -4, 5, -6]
doAssert nums.signedMinAbs() == -1
doAssert nums.signedMaxAbs() == -6
```
This is, of course, not recommended in current vanilla Nim, as `mapIt`/`map` always returns a sequence (allocates), but zero_funcional or slicerator should give a proper code for this.

ZoomRmc (approved):

maths/abs.nim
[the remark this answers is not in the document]
dlesnoff: Sure. I consider the conversation resolved.

dlesnoff:
I added empty lines at end of files as requested by @Panquesito7 in another PR.

maths/allocation_number.nim
dlesnoff: I removed this « algorithm ». It was not clear what it was about exactly.

dlesnoff:
@pietroppeter @ZoomRmc I just need one approving review to merge this old pull request.

maths/aliquot_sum.nim
```diff
@@ -0,0 +1,29 @@
...
+## all proper divisors of n, that is, all divisors of n other than n itself.
+## https://en.wikipedia.org/wiki/Aliquot_sum
+{.push raises: [].}
+
+runnableExamples:
+  import std/strformat
+  for number in [12, 100]:
+    echo fmt"The sum of all the proper divisors of {number} is {aliquotSum(number)}"
+
+func aliquotSum(number: Positive): Natural {.raises: [ValueError].} =
```
ZoomRmc, line 12:
```suggestion
  const expected = [16, 117]
  for i, number in [12, 100].pairs():
    let sum = aliquotSum(number)
    assert sum == expected[i]
    echo fmt"The sum of all the proper divisors of {number} is {sum}"

func aliquotSum*(number: Positive): Natural {.raises: [ValueError].} =
```

Module-level `runnableExamples` require the procs to be exported. And since I'm proposing a change, also added a testing logic. If there's a RE block, why not use it for additional testing?

maths/bitwise_addition.nim
```diff
@@ -0,0 +1,38 @@
...
+## Illustrate how to implement addition of integers using bitwise operations
+## See https://en.wikipedia.org/wiki/Bitwise_operation#Applications
+runnableExamples:
+  import std/strformat
+  var
+    a = 5
+    b = 6
+  echo fmt"The sum of {a} and {b} is {add(a,b)}"
+
+func add(first: int, second: int): int =
```
ZoomRmc, line 11:
```suggestion
func add*(first: int, second: int): int =
```

ZoomRmc (changes requested):

ZoomRmc (changes requested):
`abs.nim` requires a few changes to make `runnableExamples` compile:
* export all used procs
* replace `absMaxSort` with the updated name `signedMaxAbs`
You can also add a `signedMinAbs` test there, while you're editing it.

BTW, IIRC, `runnableExamples` are compiled as `debug` by default (or always?), so a regular `assert` works in them just fine. Nothing wrong with `doAssert` too, of course. Just a tip.

maths/aliquot_sum.nim
dlesnoff: RE blocks should not be used for testing, but by expliciting the expected values the example improved.

maths/abs.nim
```diff
@@ -0,0 +1,120 @@
...
+        signedMinAbs(@[0, 5, 1, 11]) == 0
+        signedMinAbs(@[3, -2, 1, -4, 5, -6]) == 1
+        signedMinAbs(@[3, -2, -1, -4, 5, -6]) == -1
+
+
+    test "Among two minimal elements, the first one is returned":
+      check signedMinAbs(@[3, -2, 1, -4, 5, -6, -1]) == 1
+
+  suite "Check `signedMaxAbs`":
+    test "Check `signedMaxAbs`":
```
dlesnoff, line 113:
```suggestion
  suite "Check signedMaxAbs":
    test "Check signedMaxAbs":
```

maths/abs.nim
```diff
@@ -0,0 +1,120 @@
...
+      check:
+        signedMaxAbs(@[3, -2, 1, -4, 5, -6]) == -6
+        signedMaxAbs(@[0, 5, 1, 11]) == 11
+
+    test "`signedMaxAbs` on empty sequence raises ValueError":
```
dlesnoff, line 118:
```suggestion
    test "signedMaxAbs on empty sequence raises ValueError":
```

maths/abs.nim
```diff
@@ -0,0 +1,120 @@
...
+        absMax(@[0, 5, 1, 11]) == 11
+        absMax(@[3, -10, -2]) == 10
+        absMax(@[-1, 2, -3]) == 3
+
+    test "`absMax` on empty sequence raises ValueError":
+      doAssertRaises(ValueError):
+        discard absMax(@[])
+
+  suite "Check `signedMinAbs`":
+    test "Check `signedMinAbs`":
```
dlesnoff, line 102:
```suggestion
  suite "Check signedMinAbs":
    test "Check signedMinAbs":
```

maths/abs.nim
```diff
@@ -0,0 +1,120 @@
...
+    let nAbs = absVal(n)
+    if nAbs > maxAbs: (max, maxAbs) = (n, nAbs)
+  max
+
+when isMainModule:
+  import std/[unittest, random]
+  randomize()
+
+  suite "Check `absVal`":
+    test "Check `absVal`":
```
dlesnoff, line 68:
```suggestion
  suite "Check absVal":
    test "Check absVal":
```

maths/aliquot_sum.nim
ZoomRmc: That's just a suggestion, Personally, I don't see anything wrong with it, as long as the code block still fulfils its explanatory functions. Personally, I had some bugs not covered by test cases found by checking in RE blocks occasionally.

maths/abs.nim
```diff
@@ -0,0 +1,120 @@
...
+      check:
+        absVal(11.2) == 11.2
+        absVal(5) == 5
+        absVal(-5.1) == 5.1
+        absVal(-5) == absVal(5)
+        absVal(0) == 0
+
+  suite "Check `absMin`":
+    test "Check `absMin`":
```
dlesnoff, line 77:
```suggestion
  suite "Check absMin":
    test "Check absMin":
```

maths/abs.nim
```diff
@@ -0,0 +1,120 @@
...
+        absMin([-1, 2, -3]) == 1
+        absMin([0, 5, 1, 11]) == 0
+        absMin([3, -10, -2]) == 2
+
+    test "`absMin` on empty sequence raises ValueError":
+      doAssertRaises(ValueError):
+        discard absMin(@[])
+
+  suite "Check `absMax`":
+    test "Check `absMax`":
```
dlesnoff, line 91:
```suggestion
    test "absMin on empty sequence raises ValueError":
      doAssertRaises(ValueError):
        discard absMin(@[])

  suite "Check absMax":
    test "Check absMax":
```

maths/aliquot_sum.nim
```diff
@@ -0,0 +1,32 @@
...
+  ## Example: aliquotSum(12) = 1 + 2 + 3 + 4 + 6 = 16
+  result = 0
+  for divisor in 1 .. (number div 2):
+    if number mod divisor == 0:
+      result += divisor
+
+when isMainModule:
+  import std/unittest
+  suite "Check `aliquotSum`":
+    test "`aliquotSum` on small values":
```
dlesnoff, line 26:
```suggestion
  suite "Check aliquotSum":
    test "aliquotSum on small values":
```

dlesnoff (commented):
Remove quotation marks for MD inline code (without opening my editor ~~since I am lazy~~).

ZoomRmc:
I think the RE block in `abs.nim` is the last thing to fix, and then it looks ready to merge!

dlesnoff:
`nim -ax r aliquot_sum.nim` made me realize there was a supplementary ValueError declaration to the exception system.
Thanks for your suggestion changes, I finally opened my editor and applied them ;)

ZoomRmc (approved):
Nicely done!

It was interesting to see the code evolve to conform to the guidelines :)

ZoomRmc (approved):

dlesnoff:
@pietroppeter You requested changes, that might block the PR. If you are still around, can you accept the PR, to see if that's what is blocking the PR?
@Panquesito7 I think that you dimissed the review of @ZoomRmc and that merging requires your approval. (At least, as a code owner, you can resolve the situation no matter the old requested changes from @pietroppeter ).

pietroppeter (approved):

dlesnoff:
Looks like it requires @Panquesito7 's approval nonetheless.

Panquesito7:
> Looks like it requires @Panquesito7 's approval nonetheless.

It requires approval from someone with write+ access.
There are not many maintainers currently. If someone wants to apply: #9

Panquesito7 (approved):

dlesnoff:
Thanks @Panquesito7 !

PR #5, opened 2023-03-16
Add the utility to automatically build `DIRECTORY.md`

ZoomRmc:
This can help keeping the DIRECTORY.md up-to-date.
If accepted, CONTRIBUTION.md needs to mention this utility and note we expect the first line of each file to have only the algorithm's title.

Usage:
`nim r directory.nim > DIRECTORY.md`

I wanted to make it work at CT, but it's too much work for no benefit, you need the compiler anyway. This could be a bash script, but it's logical the one working on this repo *has* the compiler.

Implementation details are in the source.
Currently it's pretty dumb and supports only one-level deep directory. No args, no frills.

Current output:
```markdown
<!--- DO NOT EDIT: This file is automatically generated by `directory.nim` -->

# The Algorithms — Nim: Directory

## Strings
  * [Check Anagram](strings/check_anagram.nim)

```
### Edit: On inferring the canonical algorithm name
My reasoning for it trying to extract the title from the source and not just the file name is that for some algorithms the Title Case is not correct. Some examples from the Python directory:
- Count **1S** Brian Kernighan Method
- A1**Z**26
- Vector3 For **2D** Rendering
- Transformations **2D**
- **3N** Plus 1
- Points Are Collinear *3D*
- Cramers Rule 2**X**2

Another examples are "Aho-**C**orasick" (for which the canonical name strangely omits the dash) or "A-Simple-**GAN**".

An alternative approach would be a hardcoded dictionary. Not really sure how the site decides how to lump the solutions together, but they have [the list of aliases](https://github.com/TheAlgorithms/website/blob/main/lib/aliases.ts).


dlesnoff:
> I wanted to make it work at CT, but it's too much work for no benefit

I neither see why it would be interesting to make it work at CT. I also strongly disapprove of the use of a bash script since people here are probably more familiar with Nim. It will be easier to update.

> [The program] supports only one-level deep directory.

We may safely assume the file names are named following the snake_case convention. We will probably put files under sub-directories but we may update the script when the need appears in a subsequent PR.

The script assumes our files will have a title as a header under a comment.
I do not want to enforce this rule *a priori*. It feels redundant with the file's name. I prefer stripping the file's title, writing it in Camel Case and replacing the underscores with a space.

You probably want to center this file header (instead of the Markdown's limited h1 style) using :
```html
<h1 style="text-align: center;">The Algorithms - Nim: Directory Hierarchy</h1>
```
"Directory" being a filename convention, we may use a title more explicit like Directory Hierarchy.

ZoomRmc:
> > [The program] supports only one-level deep directory.
> 
> We may safely assume the file names are named following the snake_case convention. We will probably put files under sub-directories but we may update the script when the need appears in a subsequent PR.

This was my thinking exactly.

> The script assumes our files will have a title as a header under a comment. I do not want to enforce this rule _a priori_. It feels redundant with the file's name. I prefer stripping the file's title, writing it in Camel Case and replacing the underscores with a space.

This is just a suggestion. Let's see if anyone else has an opinion on this and we can remove this then or not.

> You probably want to center this file header (instead of the Markdown's limited h1 style) using :
> 
> ```
> <h1 style="text-align: center;">The Algorithms - Nim: Directory Hierarchy</h1>
> ```

TBH, I'd prefer using the basic markdown as much as possible (with the exception of the root Readmes, which usually ask for some eye-candy).

directory.nim
```diff
@@ -0,0 +1,83 @@
+## Build the contents of `directory.md` for The Algorithms/Nim repository
+##
+## Usage: `nim r directory.nim > DIRECTORY.md`
```
pietroppeter, line 3: We probably could write directly to DIRECTORY.md (collect every output in a string and use writeFile), since it is unlikely it will be used in other ways.

pietroppeter (approved):
Just added a suggestion but looks good to me anyway, thanks!

pietroppeter:
> > The script assumes our files will have a title as a header under a comment. I do not want to enforce this rule a priori. It feels redundant with the file's name. I prefer stripping the file's title, writing it in Camel Case and replacing the underscores with a space.
> This is just a suggestion. Let's see if anyone else has an opinion on this and we can remove this then or not.

I think the script assumes that, if the first line is a documentation comment, then it should be used as a title, which seems a reasonable requirement. There is still the fallback mechanism of using filename.
Personally slightly in favor of keeping as is, but I would not object to a change.

directory.nim
ZoomRmc: I usually prefer when overwriting the file requires a deliberate action on user's/script's part, this achieves the goal without needing to parse arguments. Also, having the ability to pipe to a pager for preview is also nice.

dlesnoff:
If the filename is used as a fallback, that is fine by me.

dlesnoff (dismissed):

directory.nim
pietroppeter: Ok, in Windows I sometime run into issues with piping and not sure if I will need to add verbosity argument to avoid having compiler output in the file.

ZoomRmc: Hope the compiler prints to stderr, at least on `r`. :)
Well, output redirection to file never failed me on Windows, the output in the OP is from Windows too.

Panquesito7 (commented):
I'd create a `scripts` folder for this, or add it in the `.github` folder.

Also, there's already a (WIP?) directory workflow from the [scripts](https://github.com/TheAlgorithms/scripts).
However, I think using the repository's implementation is fine if wanted, so feel free to keep it as-is. 🙂

ZoomRmc:
I suggest we keep this for manual use until we get the proper `workflow` going when we can run the [script](https://github.com/TheAlgorithms/scripts) automatically. I wouldn't mind if this repo's workflow used this version instead, of course. Is it NiH syndrom or dogfeeding?

Having a script right near the .md file increases visibility and that's always good. But that's not for me to decide.

dlesnoff:
> I suggest we keep this for manual use until we get the proper workflow going when we can run the [script](https://github.com/TheAlgorithms/scripts) automatically.

There is a problem that only occurs when running it manually.
If our repository contains incomplete code sources for other algorithms, the DIRECTORY.md will list them even though they are not part of the Pull Request. It requires thus an extra care from us, before merging the source code.

ZoomRmc:
Well, there's an assumption the work is done in branches and the tool runs just before a final commit, so `git diff` should bring up any issues.

pietroppeter (dismissed):
I think this is good and useful, thanks!

the only pending comment I see is this remark by @Panquesito7:
> I'd create a scripts folder for this, or add it in the .github folder.
we could move it in either of those folders (I would probably go with the scripts folder), but this could also be done later. Later anyway we should mention this script in the readme (but we need to merge readme first, so let's just merge this also.

Panquesito7 (approved):
I'm no Nim expert, however, I've seen that you folks already checked it out.
Thanks for your contributions! 🚀

Would you like to work on another PR to automatically build the `DIRECTORY.md` file with a GitHub Actions workflow? 🙂

ZoomRmc:
> Would you like to work on another PR to automatically build the `DIRECTORY.md` file with a GitHub Actions workflow? slightly_smiling_face

I think for now the repo will do alright with filling it semi-manually. If the rate of PRs increases we'll see how people treat the `DIRECTORY.md` and if automated commits can create even more confusion.

PR #7, opened 2023-04-24
Add Catalan numbers

dlesnoff:
I have implemented all formulas to compute Catalan Numbers and have given a compile-time table computation.
I have also implemented an iterator version.
It might be too much, please let me know what you want to keep.

For comparison, I link below the existing implementations:
Math version: https://the-algorithms.com/algorithm/catalan-number
Dynamic programming: https://the-algorithms.com/algorithm/catalan-numbers
On Rosetta Code: https://rosettacode.org/wiki/Catalan_numbers

### DIFF
--- a/dynamic_programming/catalan_numbers.nim
+++ b/dynamic_programming/catalan_numbers.nim
@@ -1,0 +1,105 @@
+# Catalan Numbers
+#[  
+    The Catalan numbers are a sequence of natural numbers that occur in the 
+    most large set of combinatorial problems.
+    For example, it describes:
+    - the number of ways to parenthesize a product of n factors
+    - the number of ways to form a binary search tree with n+1 leaves
+    - the number of Dyck paths of length 2n
+    - the number of ways to triangulate a convex polygon with n+2 sides
+    References:
+    https://en.wikipedia.org/wiki/Catalan_number
+    https://oeis.org/A000108
+]#
+import math
+{.push raises: [].}
+
+func catalanNumbersRecursive(index: Natural): Positive =
+  ## Returns the index-th Catalan number recursively.
+  ## As Nim does not make automatic memoization, this function is not
+  ## efficient.
+  if index == 0:
+    return 1
+  for i in 0 ..< index:
+    result += catalan_numbers_recursive(i) *
+     catalan_numbers_recursive(index - i - 1)
+
+func catalanNumbersRecursive2(index: Natural): Positive =
+  if index == 0:
+    return 1
+  else:
+    2 * (2 * index - 1) * catalanNumbersRecursive2(index - 1) div (index + 1)
+
+func catalanNumbers(index_limit: Natural): seq[Positive] =
+  ## Returns all Catalan numbers up to the index-th number iteratively.
+  var catalan_numbers = newSeq[Positive](index_limit + 1)
+  catalan_numbers[0] = 1
+  for i in 1 .. index_limit:
+    for j in 0 ..< i:
+      catalan_numbers[i] += catalan_numbers[j] * catalan_numbers[i - j - 1]
+  catalan_numbers
+
+func catalanNumbers2(index: Natural): Positive =
+  ## Returns the index-th Catalan number iteratively with a second formula.
+  ## This function is not efficient.
+  binom(2 * index, index) div (index + 1)
+
+iterator catalanNumbersIt(index: Natural): Positive =
+  ## Iterates over all Catalan numbers up to the index-th number.
+  var catalan_numbers_lst = newSeq[Positive](index + 1)
+  catalan_numbers_lst[0] = 1
+  yield 1
+  for i in 1 .. index:
+    for j in 0 ..< i:
+      catalan_numbers_lst[i] += catalan_numbers_lst[j] * catalan_numbers_lst[i - j - 1]
+    yield catalan_numbers_lst[i]
+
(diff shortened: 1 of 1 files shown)


dynamic_programming/catalan_numbers.nim
```diff
@@ -0,0 +1,105 @@
...
+    
+    test "The sixteen first Catalan numbers with iterator":
+      let n = 15
+      var i = 0
+      for catalan_number in catalan_numbers_it(n):
+        check catalan_number == expectedResult[i]
+        inc i
+    
+    test "Uses compile-time table":
+      check catalan_number(36) == Positive(3116285494907301262)
```
Panquesito7, line 105: Missing an endline. Without this endline, there could be issues with some IDEs and configurations.
```suggestion
      check catalan_number(36) == Positive(3116285494907301262)
      
```

Panquesito7 (changes requested):

dynamic_programming/catalan_numbers.nim
dlesnoff: Done.

Panquesito7 (approved):
I'm not proficient with Nim, but this looks good. Thanks. 🚀
Will wait until other people check it out, if not, I'll merge it in the next few days.

dynamic_programming/catalan_numbers.nim
```diff
@@ -0,0 +1,105 @@
...
+
+func catalanNumbersRecursive(index: Natural): Positive =
+  ## Returns the index-th Catalan number recursively.
+  ## As Nim does not make automatic memoization, this function is not
+  ## efficient.
+  if index == 0:
+    return 1
+  for i in 0 ..< index:
+    result += catalan_numbers_recursive(i) *
+     catalan_numbers_recursive(index - i - 1)
```
ZoomRmc, line 25:
```suggestion
  if index < 2:
    return 1
  for i in 0 ..< index:
    result += catalanNumbersRecursive(i) *
     catalanNumbersRecursive(index - i - 1)
```

dynamic_programming/catalan_numbers.nim
```diff
@@ -0,0 +1,105 @@
...
+
+func catalanNumbersRecursive2(index: Natural): Positive =
+  if index == 0:
```
ZoomRmc, line 28:
```suggestion
  if index < 2:
```

dynamic_programming/catalan_numbers.nim
```diff
@@ -0,0 +1,105 @@
...
+    2 * (2 * index - 1) * catalanNumbersRecursive2(index - 1) div (index + 1)
+
+func catalanNumbers(index_limit: Natural): seq[Positive] =
+  ## Returns all Catalan numbers up to the index-th number iteratively.
+  var catalan_numbers = newSeq[Positive](index_limit + 1)
+  catalan_numbers[0] = 1
+  for i in 1 .. index_limit:
+    for j in 0 ..< i:
+      catalan_numbers[i] += catalan_numbers[j] * catalan_numbers[i - j - 1]
+  catalan_numbers
```
ZoomRmc, line 40: I'm not a fan of `result`s but since we're using them anyway... Not reusing an identifier is a bonus in this case.
```suggestion
func catalanNumbers(indexLimit: Natural): seq[Positive] {.noinit.} =
  ## Returns all Catalan numbers up to the index-th number iteratively.
  result = newSeq[Positive](indexLimit + 1)
  result[0] = 1
  for i in 1 .. indexLimit:
    for j in 0 ..< i:
      result[i] += result[j] * result[i - j - 1]
```

dynamic_programming/catalan_numbers.nim
```diff
@@ -0,0 +1,105 @@
...
+
+iterator catalanNumbersIt(index: Natural): Positive =
+  ## Iterates over all Catalan numbers up to the index-th number.
+  var catalan_numbers_lst = newSeq[Positive](index + 1)
+  catalan_numbers_lst[0] = 1
+  yield 1
+  for i in 1 .. index:
+    for j in 0 ..< i:
+      catalan_numbers_lst[i] += catalan_numbers_lst[j] * catalan_numbers_lst[i - j - 1]
+    yield catalan_numbers_lst[i]
```
ZoomRmc, line 55: Multiple yields in regular Nim iterators should be avoided as they can lead to considerable code duplication. Here it's doable and is even shorter.

```suggestion
  var catalanNumbers = newSeq[Positive](index + 1)
  catalanNumbers[0] = 1
  for i in 0 .. index:
    for j in 0 ..< i:
      catalanNumbers[i] += catalanNumbers[j] * catalanNumbers[i - j - 1]
    yield catalanNumbers[i]
```

ZoomRmc (changes requested):

Panquesito7:
Thanks for your review, @ZoomRmc! 🚀

ZoomRmc (approved):

ZoomRmc:
Thanks! Keep on doing the good work, @dlesnoff!

dlesnoff:
I extensively tested the code.
I added limits for the different methods.
To compute directly a number for a specific index, the other formulas are probably better, but they overflow due to an intermediate computation. I added these overflow limits as consts.

I fix a defect in the use of the compile time table.

I made indexes consistent with the comment, and stop the computations before the input parameter.

Panquesito7 (approved):

PR #8, opened 2023-04-24
Viterbi

dlesnoff:
The Viterbi algorithm provides the most probable sequence of states of a hidden Markov model given a sequence of observations.


Panquesito7 (approved):
Not an expert in Nim, but this looks good. Thanks. 🚀
Will wait until someone else reviews it, if not, I'll merge it in the next few days.

PR #12, opened 2023-04-29
Add insertion sort

dlesnoff:


Panquesito7 (approved):
Will merge in a few days if there are no objections. Thanks! 🚀

dlesnoff:
I will put all sorting algorithms I made in an upcoming PR, so that we can have a discussion on some approaches common to the different algorithms.

PR #13, opened 2023-04-29
Add bubble sort

dlesnoff:
Add bubble sort
Bubble sort is probably amongst the sorting algorithms both the easiest 
to memorize as well as the slowest. I provide four implementations.
The first implementation is a direct implementation of the algorithm 
principle.
The second implementation uses a boolean flag to break the outer loop 
early.
The third implementation implements the same idea with a while loop 
instead. This eliminates the need for an early return.
The fourth implementation uses a second optimization taken from 
Wikipedia's 
pseudocode.

### DIFF
--- a/sorts/bubble_sort.nim
+++ b/sorts/bubble_sort.nim
@@ -1,0 +1,79 @@
+# Bubble sort
+#[
+Bubble sort, sometimes referred to as sinking sort, is a simple sorting algorithm that repeatedly steps through the input list element by element, comparing the current element with the one after it, swapping their values if needed. These passes through the list are repeated until no swaps had to be performed during a pass, meaning that the list has become fully sorted.
+# https://en.wikipedia.org/wiki/Bubble_sort
+]#
+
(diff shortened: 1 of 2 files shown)


dlesnoff:
Sorry I have some discrepancies in my git log that I do not understand, I am trying to fix that, hence the dummy commits and the numerous force-push.

sorts/bubble_sort.nim
```diff
@@ -0,0 +1,79 @@
...
+      check testSort(bubbleSort[float])
+      check testSort(bubbleSortOpt1[float])
+      check testSort(bubbleSortWhileLoop[float])
+      check testSort(bubbleSortOpt2[float])
+    test "Sort characters":
+      check testSort(bubbleSort[char])
+      check testSort(bubbleSortOpt1[char])
+      check testSort(bubbleSortWhileLoop[char])
+      check testSort(bubbleSortOpt2[char])
+
```
Panquesito7, line 79:
```suggestion
```

Panquesito7 (approved):

Issue #16, opened 2023-05-14
Add a GitHub workflow that checks the code (Nim linter)


Panquesito7:

dlesnoff:
Closed by error. Check_code_format is not a linter.

PR #19, opened 2023-05-30
chore: add `check_code_format` workflow

vil02:
This PR adds a workflow running the command `nim prettyfy` on the entire project.

The added workflow is executed on every push to master and pull reques

It turned out, that there are already two files, which are not correctly formatted: c6acd826bc39f97b269f4e1a7d7fc4bcf8e3f062.

The shell script could be removed thanks to #21.

### DIFF
--- a/.github/workflows/run_nimpretty.yml
+++ b/.github/workflows/run_nimpretty.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - master
+      - main
   pull_request:
 
 jobs:
--- a/.github/workflows/run_nimpretty.yml
+++ b/.github/workflows/run_nimpretty.yml
@@ -16,6 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.head.ref }}
           fetch-depth: 0
 
       - uses: jiro4989/setup-nim-action@v1
--- a/.github/workflows/run_nimpretty.yml
+++ b/.github/workflows/run_nimpretty.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Fail if needs reformatting
         run: |
-          if [[ $(git status --porcelain=v1 | wc -l) ]]; then
+          if [[ $(git status --porcelain) ]]; then
              echo "please reformat these files with nimpretty:"
              git status --porcelain=v1
              exit 1
--- a/.github/workflows/run_nimpretty.yml
+++ b/.github/workflows/run_nimpretty.yml
@@ -5,6 +5,9 @@ name: run_nimpretty
 on:
   workflow_dispatch:
   push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   format_code_with_nimpretty:
@@ -21,17 +24,11 @@ jobs:
         run: |
           ./run_nimpretty.sh
 
-      - name: Configure git
+      - name: Fail if needs reformatting
         run: |
-          git config --global user.name github-action
-          git config --global user.email '${GITHUB_ACTOR}@users.noreply.github.com'
-
-      - name: Commit changes
-        run: |
-          if [[ `git status --porcelain` ]]; then
-            git commit -am "style: formatting with nimpretty"
-            git push origin HEAD:$GITHUB_REF
-          else
-            echo "NO CHANGES DETECTED"
+          if [[ $(git status --porcelain=v1 | wc -l) ]]; then
+             echo "please reformat these files with nimpretty:"
+             git status --porcelain=v1
+             exit 1
           fi
 ...
--- a/.github/workflows/run_nimpretty.yml
+++ b/.github/workflows/run_nimpretty.yml
@@ -29,6 +29,9 @@ jobs:
           git config --global user.name github-action
           git config --global user.email '${GITHUB_ACTOR}@users.noreply.github.com'
 
+      - name: display changes
+        run: git status
+
       - name: Commit changes
         run: |
           if [[ `git status --porcelain` ]]; then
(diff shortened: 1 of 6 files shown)


dlesnoff:
I am on a rush, I will test it during the week-end.

I made a lot of spaces on empty lines :(

Very nice first contribution. I guess this fixes issue #16 .

.github/workflows/run_nimpretty.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Commit changes
+        run: |
+          if [[ `git status --porcelain` ]]; then
+            git commit -am "style: formatting with nimpretty"
+            git push origin HEAD:$GITHUB_REF
+          else
+            echo "NO CHANGES DETECTED"
+          fi
+...
```
Panquesito7, line 37: I don't think these dots are intended to be here.
```suggestion
```

.github/workflows/run_nimpretty.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Configure git
+        run: |
+          git config --global user.name github-action
+          git config --global user.email '${GITHUB_ACTOR}@users.noreply.github.com'
+
+      - name: Commit changes
+        run: |
+          if [[ `git status --porcelain` ]]; then
+            git commit -am "style: formatting with nimpretty"
+            git push origin HEAD:$GITHUB_REF
```
Panquesito7, line 33: Creating and pushing a commit will not be allowed due to branch protection. Giving GitHub Actions full permissions is not recommended due to a few [number of reasons](https://github.com/orgs/community/discussions/25305#discussioncomment-3247401).
I recommend creating a PR and letting the maintainers review the PR created by the workflow, or we can merge it directly as well.

Also, I think it would be great if this could run on every `push` rather than workflow dispatch. 🙂

```suggestion
            git checkout -b directory-${{ github.sha }} # An unique branch name is recommended for each commit made on the repository.
            gh pr create --base ${GITHUB_REF##*/} --head directory-${{ github.sha }} --title 'style: formatting with `nimpretty`' --body '`Files have been modified with `nimpretty` (see diff. for changes)..'
```

Panquesito7 (changes requested):

.github/workflows/run_nimpretty.yml
vil02: I did in on purpose cf. [yaml docs](https://yaml.org/spec/1.1/#document%20boundary%20marker/).

.github/workflows/run_nimpretty.yml
vil02: Then maybe the workflow should simply _fail_, when the code if not formatted correctly. Having pull requests for almost every push might be impractical.

The suggested workflow can be triggered both manually (`workflow_dispatch:`) or by push.

vil02:
> Very nice first contribution. I guess this fixes issue #16 .

Nimpretty, as far as I can tell, is a code formatter - it does not perform any static analysis.

.github/workflows/run_nimpretty.yml
Panquesito7: > Then maybe the workflow should simply _fail_, when the code if not formatted correctly. Having pull requests for almost every push might be impractical. and automatic merging will also require write access.
> 
> The suggested workflow can be triggered both manually (`workflow_dispatch:`) or by push.

Write access in the repository settings shouldn't harm, and in the suggestion above I'm just creating a PR to the repository.
If you don't want on every push, you can just push changes to a certain branch, maybe. 🤔

Panquesito7 (approved):
Looks great! Thanks. 🚀

.github/workflows/run_nimpretty.yml
```diff
@@ -0,0 +1,35 @@
+---
+name: run_nimpretty
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
```
dlesnoff, line 9:
```suggestion
      - main
```

dlesnoff (changes requested):

ZoomRmc:
I propose replacing the bash script with a task from #21.

.github/workflows/run_nimpretty.yml
vil02: Thanks for spotting that.

.github/workflows/run_nimpretty.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  format_code_with_nimpretty:
+    name: format code with nimpretty
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: jiro4989/setup-nim-action@v1
```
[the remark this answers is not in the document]
vil02: Here and in #20 there is:
- `jiro4989/setup-nim-action@v1` (cf. [nim_test.yml](https://github.com/TheAlgorithms/Nim/blob/19f49b4f7af0d8d2a952b277edddac90c991bc26/.github/workflows/nim_test.yml#L26), [`run_nimpretty.yml`](https://github.com/TheAlgorithms/Nim/blob/8e4b03c8105094959dbdab3f7591e3a64dbca060/.github/workflows/run_nimpretty.yml#L21))
- 'actions/checkout@v3' (cf. [nim_test.yml](https://github.com/TheAlgorithms/Nim/blob/19f49b4f7af0d8d2a952b277edddac90c991bc26/.github/workflows/nim_test.yml#L24), [`run_nimpretty.yml`](https://github.com/TheAlgorithms/Nim/blob/8e4b03c8105094959dbdab3f7591e3a64dbca060/.github/workflows/run_nimpretty.yml#L17))

run_nimpretty.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+find . -name "*.nim" -exec nimpretty {} +
```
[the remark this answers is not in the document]
ZoomRmc: There's no hard style requirements for `scripts` and the contents are supposed to be low-churn. No point reformatting them on each CI run.

dlesnoff (approved):

PR #20, opened 2023-06-01
chore: add `nim_test.yml`

vil02:
This PR adds a very simple CI for this repo (and fixes some bugs in the existing `*.nim` files).

Note that there are still some warnings in the existing files.

~~The future work is definitely to setup this project using `nimble`, but this also would require things like project structure change, which is way too far beyond the scope of this PR.~~

Regarding the shellscripts: #21


.github/workflows/nim_test.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  nim_test:
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+        nim-version:
+          - '1.6.12'
```
dlesnoff, line 21: Add devel 
```suggestion
          - '1.6.12'
          - 'devel'
```

ZoomRmc: Do we want to require passing tests on devel?
Current guidelines mention it only as a recommendation

ZoomRmc (changes requested):
Looks good, besides the proposed change and this:
I propose replacing the Bash scripts in this PR with a nimscript task from #21.

.github/workflows/nim_test.yml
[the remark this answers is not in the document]
vil02: `stable` changes over time. My suggestion is to add new versions when needed (that is why I have introduced the `strategy/matrix`).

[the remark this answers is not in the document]
ZoomRmc: > stable changes over time.

Exactly. Isn't it the whole point of building everything automatically?

Following stable here provides early warning and ensures there's no big migration event ever.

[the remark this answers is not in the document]
vil02: Imagine you change some file and make a PR. And at the same time `stable` changes. It might turn out that on your PR suddenly many other, untouched files will fail.

I think it might be easier to simply add a new version from time to time.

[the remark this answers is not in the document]
vil02: I think it might be a good idea adding another workflow doing a build every week or month with `stable` version. Something like:
```yaml
---
name: nim_stable_test

# yamllint disable-line rule:truthy
on:
  workflow_dispatch:
  cron: '3 1 * * 1'

jobs:
  nim_test:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v3

      - uses: jiro4989/setup-nim-action@v1

      - name: Display nim version:
        run: |
          nim --version

      - name: Build and test
        run: |
          ./check_all_nim_files.sh
...
```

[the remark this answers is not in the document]
ZoomRmc: > Imagine you change some file and make a PR. And at the same time stable changes. It might turn out that on your PR suddenly many other, untouched files will fail.

If other packages fail, it's not a problem, people can notice and start upgrading them right away.

Automation should help manual reviewing, not replace it. Checking PRs on stable helps ensuring the contents of the PR actually build, and do it while the project is in the attention scope of the contributor, so they can make the necessary fixes.

This repo is not some critical piece of software which needs to be automatically deployed from the pipeline, there's no need to avoid possible minor breakage here or there at all costs. Moreover, stable is not as volatile, breaking changes are usually added for a reason.

dlesnoff:
Sorry I meant checking against devel too.
I would like to know if some code works under devel but does not work with the stable version.

dlesnoff:
I do not want to setup a Nimble/Atlas project file. I have had many problems switching between developer and user versions of the repositories. I am confused by the recent changes brought to Nimble that made it quite unstable. ) The imposed Nimble directory structure will probably conflict with the directory structure of the Algorithms repository.

Nimble is not really necessary. What advantages does it really bring ?
- Tasks in Nimble can be written in a classical .nims file.
- Dependencies handling ? We do not want dependencies.

.github/workflows/nim_test.yml
vil02: What do you think about having an additional workflow similar to one [here](https://github.com/TheAlgorithms/Nim/pull/20#discussion_r1213929735) running both `stable` and `devel`? Then there would be a cyclic check if the code still builds with the newest versions, but would not possibly spoil some PRs.

ZoomRmc: I don't mind having it, depends on what will trigger it.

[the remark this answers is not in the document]
dlesnoff: I am in favor of using stable now and fall back to manual versioning if we encounter problems.
In nim-lang/bigints, we removed automatic versioning in https://github.com/nim-lang/bigints/pull/97 to explicitly state the versions supported.

Here, we do not need backwards compatibility.

dlesnoff: I want to check tests on devel, not require passing tests.
A devel is a future stable version, it helps knowing the changes we will have to bring.

check_all_nim_files.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
dlesnoff, line 1: See #21. I prefer Nim Script tasks which are more cross-platform and easier to maintain for Nim developers.

.github/workflows/nim_test.yml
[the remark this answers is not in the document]
ZoomRmc: The argument of tests failing on units not related to the PR at hand is solid. It all boils down to what do we want from this job.

1. If we want it to be a hard-stop on any fail ("all green checks -> go!"), than it has to not fail due to underlying compiler/stdlib updates (even though the probability of it happening is not that great) and hard-coding some version is necessary.
2. If we want it to be just a thing to consider, an assistance for manual reviews, than `stable` is definitely better.

Having thought about it a bit more, I don't think I have a strong opinion on the matter now.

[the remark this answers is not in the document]
vil02: @ZoomRmc I would suggest 1 (think about the hacktoberfest madness) and the cyclically running workflow with `stable` and `devel`.

ZoomRmc (dismissed):

.github/workflows/nim_test.yml
```diff
@@ -0,0 +1,33 @@
...
+  pull_request:
+
+jobs:
+  nim_test:
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
```
Panquesito7, line 19: Why not use `ubuntu-latest` directly?

.github/workflows/nim_test.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/checkout@v3
+
+      - uses: jiro4989/setup-nim-action@v1
+        with:
+          nim-version: ${{matrix.nim-version}}
+
+      - name: Build and test
+        run: |
+          nim test
+...
```
Panquesito7, line 33: I don't think these dots should be here.
```suggestion
```

.github/workflows/nim_test_cyclic.yml
```diff
@@ -0,0 +1,36 @@
...
+          nim-version: ${{matrix.nim-version}}
+
+      - name: Display nim version
+        run: |
+          nim --version
+
+      - name: Build and test
+        run: |
+          nim test
+...
```
Panquesito7, line 36: Same here.
```suggestion
```

Panquesito7 (changes requested):

.github/workflows/nim_test.yml
vil02: To make it easier in the future to introduce other operating systems.

.github/workflows/nim_test.yml
vil02: They are already [here](https://github.com/TheAlgorithms/Nim/blob/a2fc1bd53dcd8a320cc67074d35c02f2b73ca209/.github/workflows/check_code_format.yml#L35). I think it is good to be consistent.

.github/workflows/nim_test_cyclic.yml
vil02: As [above](https://github.com/TheAlgorithms/Nim/pull/20#discussion_r1215297583).

nim.cfg
```diff
@@ -0,0 +1,2 @@
+# Nim compilation defaults for the whole repository
+--gc: arc
```
dlesnoff, line 2: Why did you put these in a new nim.cfg file?
I think that .nims file superseded nim.cfg files: see e.g. https://forum.nim-lang.org/t/6844#42785

config.nims
```diff
@@ -1,5 +1,3 @@
-# Nim compilation defaults for the whole repository
-#--gc: arc
```
dlesnoff, line 2: If the --gc: arc was commented, that is a merging error, please uncomment it directly in this file

nim.cfg
ZoomRmc: This is a workaround for a bug in stable nim:
https://github.com/TheAlgorithms/Nim/pull/21#issuecomment-1574393251

My bad, turns out I was on devel when I checked this.

config.nims
ZoomRmc: See above

dynamic_programming/catalan_numbers.nim
```diff
@@ -113,7 +114,7 @@ when isMainModule:
     test "We can compute up to the thirty-seventh Catalan number iteratively":
       let limit = UpperLimit
       let catalanNumbersSeq = catalanNumbers(limit)
-      for index in 0 .. limit:
+      for index in 0 ..< limit:
         check catalanNumbersSeq[index] == CatalanNumbersList[index]
```
dlesnoff, line 118: This is not a syntax issue. Have you run the tests? I am pretty sure it worked by including the upper bound.

dlesnoff (changes requested):

nim.cfg
dlesnoff: I do not understand. Had you only the error on your devel version? Please, open an issue to remember to fall back entirely to .nims configuration after Nim 2.0 release.

dynamic_programming/catalan_numbers.nim
dlesnoff: This algorithm does not have plenty of tests.
I will try to imagine some easy examples not to hard to compute by hand for better verification after this P.R.

vil02: > This is not a syntax issue. Have you run the tests? I am pretty sure it worked by including the upper bound.

- [`limit`](https://github.com/TheAlgorithms/Nim/blob/c38d3bc59bf3bb79ac6ca1d58d73d531d4d497ae/dynamic_programming/catalan_numbers.nim#L115) is set to [`36`](https://github.com/TheAlgorithms/Nim/blob/c38d3bc59bf3bb79ac6ca1d58d73d531d4d497ae/dynamic_programming/catalan_numbers.nim#L80),
- [`CatalanNumbersList`](https://github.com/TheAlgorithms/Nim/blob/c38d3bc59bf3bb79ac6ca1d58d73d531d4d497ae/dynamic_programming/catalan_numbers.nim#L85) has 36 elements, the last element is `CatalanNumbersList[35]`
- [`catalanNumbersSeq`](https://github.com/TheAlgorithms/Nim/blob/c38d3bc59bf3bb79ac6ca1d58d73d531d4d497ae/dynamic_programming/catalan_numbers.nim#L116) also has 36 elements (note `<` in [catalanNumbers](https://github.com/TheAlgorithms/Nim/blob/c38d3bc59bf3bb79ac6ca1d58d73d531d4d497ae/dynamic_programming/catalan_numbers.nim#L37)).
- without this change I get `Unhandled exception: index 36 not in 0 .. 35 [IndexDefect]` (both by executing the command `nim test` or by running this file directly e.g. `nim r catalan_numbers.nim`).

nim.cfg
ZoomRmc: On the contrary. I was on devel and missed this error. I'll open the issue when I (or someone else) identify the fixing PR for Nim to watch it landing in stable.

dlesnoff (approved):
Improves greatly the repository's state and workflow. I need Panquesito's approval to merge this.

dynamic_programming/catalan_numbers.nim
dlesnoff: I was thinking about Viterbi. Catalan Numbers does have plenty of tests, which made us notice this bug (probably a merge error).

.github/workflows/nim_test.yml
dlesnoff: Why not testing it directly against Windows and MacOS ?

vil02: Done in de0c8d2bf332623f5f78dbacd33d3822472f52de.

.github/workflows/nim_test.yml
```diff
@@ -0,0 +1,35 @@
+---
+name: nim_test
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
```
Panquesito7, line 6: Not sure we want workflow dispatch when it already runs automatically. 🤔

vil02: _Just in case_. Having this extra line does not hurt and it helps to avoid some _blank_ PR's just to run some workflow (cf. TheAlgorithms/Go#638)

ZoomRmc:
Nice work everybody!

PR #22, opened 2023-06-02
Sorting Algorithms

dlesnoff:
Add multiple sorting algorithms: insertion sort, selection sort, bubble sort, merge sort
with a file for some common tests.

I hesitate to add additional tests to illustrate the stability property of some sorts or to add the possibility to change the comparison function.
I want to keep thing simple to be as much pedagogical as possible.


DIRECTORY.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Dynamic Programming
+  * [Catalan Numbers](dynamic_programming/catalan_numbers.nim)
+  * [Viterbi](dynamic_programming/viterbi.nim)
+
+## Maths
+  * [Area](maths/area.nim)
+
+## Sorts
+  * [Bubble sort](sorts/bubble_sort.nim)
```
dlesnoff, line 13: I just noticed that capitalization of titles is not coherent.

Panquesito7: That is something related to the directory script.
Another issue should be created for that, IMO.

ZoomRmc:
Took a glance over. Looks great, I like the code/comment ratio (I'd spray TODO marks around those `add description` lines).

I think the testing scaffolding should live somewhere separately. Do you think it would be beneficial to start collecting the testing utils we can recommend to new contributors?

Also, in general, I don't like unseeded random in tests. No harm here, but they should be reproducible.

Re `if verbose` that you don't actually use, perhaps the checks can be swapped to `when defined(debug)`?

PR #30, opened 2023-06-19
Add iterative and recursive linear searches

SatinWuker:
I only added one searching algorithm to this directory, linear search, because I just want to make sure that the comments, format, and code style I did are correct and appropriate according to the contribution guidelines. If `linear_search.nim` is okay, I will continue working on other searching algorithms. @dlesnoff
And please review my code, I can't wait to make more contributions haha
Thank you!


searchings/linear_search.nim
```diff
@@ -0,0 +1,68 @@
...
+# https://www.simplilearn.com/tutorials/data-structure-tutorial/linear-search-algorithm
+]#
+## Worst case time complexity: O(N)
+## Average case time complexity: O(N)
+## Best case time complexity: O(1)
+## Space complexity: O(1)
+
+## openArray[T] is a func parameter type that accept arrays and seqs in any type
+# value is the value for matching in the array
+func linearSearch[T](arr: openArray[T], value:T): int =
```
dlesnoff, line 18: Be consistent in your spacing
```suggestion
func linearSearch[T](arr: openArray[T], value: T): int = 
```

searchings/linear_search.nim
```diff
@@ -0,0 +1,68 @@
...
+to search for an element in a data set.
+It examines each element until it finds a match,
+starting at the beginning of the data set, until the end.
+The search is finished and terminated once the target element is located.
+# https://www.simplilearn.com/tutorials/data-structure-tutorial/linear-search-algorithm
+]#
+## Worst case time complexity: O(N)
+## Average case time complexity: O(N)
+## Best case time complexity: O(1)
+## Space complexity: O(1)
```
dlesnoff, line 14: Not for the recursive algorithm. I think you can assume that you are in O(n) due to the call stack. I do not think there are any [CPS](https://en.wikipedia.org/wiki/Continuation-passing_style) automatic optimization in the Nim compiler.

searchings/linear_search.nim
```diff
@@ -0,0 +1,68 @@
...
+Linear search is the simplest but least efficient algorithm
+to search for an element in a data set.
+It examines each element until it finds a match,
+starting at the beginning of the data set, until the end.
+The search is finished and terminated once the target element is located.
+# https://www.simplilearn.com/tutorials/data-structure-tutorial/linear-search-algorithm
+]#
+## Worst case time complexity: O(N)
+## Average case time complexity: O(N)
+## Best case time complexity: O(1)
```
dlesnoff, line 13: Define above what you understand by worst, average, best, just to be sure.

searchings/linear_search.nim
```diff
@@ -0,0 +1,68 @@
...
+      check linearSearch(arr, 5) == 4
+      check recursiveLinearSearch(arr, arr.len - 1, 5) == 4
+
+    test "Search in int array matching with invalid value":
+      var arr = @[0, 3, 1, 4, 5, 6]
+
+      check linearSearch(arr, 7) == -1
+      check recursiveLinearSearch(arr, arr.len - 1, 7) == -1
+
+    test "Search in char array matching with char matching value":
```
dlesnoff, line 58:
```suggestion
    test "Search in a char array matching with char matching value":
```

dlesnoff (commented):
Thanks for this new contribution!
I made some style comments. I am going to do a second pass for the algorithm itself tonight.

dlesnoff:
I would copy other the-algorithms repositories and name the repertory searches instead.

searchings/linear_search.nim
SatinWuker: Sorry, it was an inaccuracy. No matter whether the inputs of an algorithm result in different runtime/efficiency/actual time, the time complexity does not change. It is still O(n). However What I meant by `worst case time complexity` is just the worst case the time of the algorithm can be, resulted by the input. Those two lines can actually be removed, as they are not actually helpful for users. Developers can figure out what the worst case and best case are, rightaway when they see the time complexity. I will change!

ZoomRmc:
Nicely done! A few general remarks:

* Lines starting with double octothorps are [doc comments](https://nim-lang.org/docs/docstyle.html). To work properly, they go under the thing, using the proper indentation.
* In the header the multiline comment is a regular one, it won't be included in the generated documentation.
* Since Nim is very overload heavy, following argument conventions (including naming and ordering) is more important than in many other languages. For searches, consider following the standard library and use `key` for the searched argument name. (See [`binarySearch`](https://nim-lang.org/docs/algorithm.html#binarySearch%2CopenArray%5BT%5D%2CT))
* Explaining what the common types (like openArray) is probably excessive.
* There's a `..<` for "[A, B)" (end not included) ranges.
* Use `return` only when you're modifying the control flow (early returns), for regular returns use expressions or an implicitly declared `result` (I strongly prefer the former).
  So, to return a value from a proc/block, instead of
  ```nim
  proc foo(a: T): T =
  return foo(bar)
  ``` 
  Use:
  ```nim
  proc foo(a: T): T =
  foo(bar)
  ``` 
* Consider using the module-level `runnableExamples` block for explaining the usage, comparison between implemented functions and general narration. This allows the per-funciton documentation to be more succinct and technical.
* For input index arguments, [use `Natural`](/TheAlgorithms/Nim/blob/main/CONTRIBUTING.md#types).

dlesnoff:
> For input index arguments, [use Natural](https://github.com/TheAlgorithms/Nim/blob/main/CONTRIBUTING.md#types).

Problem: the item might not be in the array, and @SatinWuker returns -1, so a Natural is not an option.
I would recommend using an Option[Natural] for the index instead.

madhavgaba (approved):

SatinWuker:
@ZoomRmc, @dlesnoff Nim is not my main language, and I rarely used Nim in my work. I am not sure how to use the Natural type as the type for the index in the for loop.
```nim
func linearSearch[T](arr: openArray[T], key: T): Option[Natural] = 
  ## key is the value for matching in the array

  for i in arr.low .. arr.high:
    if arr[i] == key:
      return some(i)
  return none(Natural) # -1 is the default index for unfound element
```
In this case, how can I make `i` a Natural-typed variable?
After fixing this, I think I completed all the requests for change.

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+  var arr1 = [0, 3, 1, 4, 5, 6]
+  doAssert linearSearch(arr1, 5) == some(Natural(4))
+  doAssert recursiveLinearSearch(arr1, arr1.high, 5) == some(Natural(4))
+
+  var arr2 = ['0', 'c', 'a', 'u', '5', '7']
+  doAssert linearSearch(arr2, '5') == some(Natural(4))
+  doAssert recursiveLinearSearch(arr2, arr2.high, '5') == some(Natural(4))
+
+
+## importing options and system for type Option and Natural
```
ZoomRmc, line 27:
```suggestion
```
It's self-explaining. If you wish to leave it, make it a regular comment (with a single hash), otherwise it will be displayed as a free-standing sentence in the documentation.

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+import std/options
+import system
```
ZoomRmc, line 29: System is implicitly imported. For importing several modules from a single source (such as stdlib), you can use the `std/[options, ...]` syntax.
```suggestion
```

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+
+func linearSearch[T](arr: openArray[T], key: T): Option[Natural] = 
+  ## key is the value for matching in the array
+  for i in arr.low .. arr.high:
+    if arr[i] == key:
+      return some(Natural(i))
+
+  return none(Natural) # -1 is the default index for unfound element
```
ZoomRmc, line 37:
```suggestion
  none(Natural) # `key` not found
```
The result is implicitly initialized to this value (the default for `Option[T]` is `none(T)`), but I agree that returning it explicitly is better for readability.

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+
+## importing options and system for type Option and Natural
+import std/options
+import system
+
+func linearSearch[T](arr: openArray[T], key: T): Option[Natural] = 
+  ## key is the value for matching in the array
+  for i in arr.low .. arr.high:
+    if arr[i] == key:
+      return some(Natural(i))
```
ZoomRmc, line 35:
```suggestion
  for i, val in arr.pairs():
    if val == key:
      return some(Natural(i))
```
[`pairs`](https://nim-lang.org/docs/iterators.html#pairs.i%2CopenArray%5BT%5D) is idiomatic Nim, no need for manual indexing.

ZoomRmc (changes requested):

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+
+  return none(Natural) # -1 is the default index for unfound element
+
+func recursiveLinearSearch[T](arr: openArray[T], idx: Natural, value: T): Option[Natural] =
+  ## Recursion is another method for linear search
+  ## we can just replace the for loop with recursion.
+  ## recursion traverses from the end of the array to the front.
+
+  ## return -1 would be invoked when the array is traversed completely
+  ## and no value is matched, or when array is empty and has a length of 0
```
dlesnoff, line 45:
```suggestion
  ## and no value is matched, or when the array is empty and has a length of 0
```
or
```suggestion
  ## and no value is matched, or when `arr` is empty and has a length of 0
```

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+
+func linearSearch[T](arr: openArray[T], key: T): Option[Natural] = 
+  ## key is the value for matching in the array
+  for i in arr.low .. arr.high:
+    if arr[i] == key:
+      return some(Natural(i))
+
+  return none(Natural) # -1 is the default index for unfound element
+
+func recursiveLinearSearch[T](arr: openArray[T], idx: Natural, value: T): Option[Natural] =
```
ZoomRmc, line 39: Consider providing a public (exported with a `*` mark) version without the `idx` arg, as it's error-prone and requiring manually calling it with the correct index is probably unnecessary.
```nim
func recursiveLinearSearch[T](arr: openArray[T], value: T, idx: Natural): Option[Natural] =
  ...

func recursiveLinearSearch*[T](arr: openArray[T], value: T): Option[Natural] =
  recursiveLinearSearch(arr, value, arr.high)
```
This will require fixing the test cases.

The other option is using a default value for the argument (`idx: Natural = arr.high`) but there's currently a bug in the codegen that forbids this option.

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+      var arr = ['0', 'c', 'a', 'u', '5', '7']
+
+      check linearSearch(arr, '5') == some(Natural(4))
+      check recursiveLinearSearch(arr, arr.high, '5') == some(Natural(4))
+
+    test "Search in a string array matching with a string matching value":
+      var arr = ["0", "c", "a", "u", "5", "7"]
+
+      check linearSearch(arr, "5") == some(Natural(4))
+      check recursiveLinearSearch(arr, arr.high, "5") == some(Natural(4))
```
dlesnoff, line 85: I would template similar tests for less redundancy.

```suggestion
  template checkLinearSearch[T](arr: openArray[T], value: T, expectedIndex: Option[Natural]) =
    check linearSearch(arr, value) == expectedIndex
    check recursiveLinearSearch(arr, arr.high, value) == expectedIndex

  suite "Linear search":
    test "Search in an empty array":
      var arr: array[0, int]

      checkLinearSearch(arr, 5, none(Natural))

    test "Search in an int array matching with a valid value":
      var arr = [0, 3, 1, 4, 5, 6]

      checkLinearSearch(arr, 5, some(Natural(4)))

    test "Search in an int array matching with an invalid value":
      var arr = [0, 3, 1, 4, 5, 6]

      checkLinearSearch(arr, 7, none(Natural))

    test "Search in a char array matching with a char matching value":
      var arr = ['0', 'c', 'a', 'u', '5', '7']

      checkLinearSearch(arr, '5', some(Natural(4)))

    test "Search in a string array matching with a string matching value":
      var arr = ["0", "c", "a", "u", "5", "7"]

      checkLinearSearch(arr, "5", some(Natural(4)))
```

searches/linear_search.nim
dlesnoff: I may go rather for the second option as the overload resolution does not take the export marker into account.

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
+## Linear Search
+## =============
+## Linear search is the simplest but least efficient algorithm
+## to search for an element in a data set.
+## It examines each element until it finds a match,
+## starting at the beginning of the data set, until the end.
```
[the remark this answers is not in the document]
dlesnoff: It is not very difficult to make the recursive version go into the same direction as the iterative version.
Currently, they do not return the same index when looking for a value present at different places in the array:
```nim
var arr = [2, 4, 5, 4]
doAssert linearSearch(arr, 4) == 1
doAssert recursiveLinearSearch(arr, 4) == 3
```

I would change the recursive algorithm and add this as a test below with 1 as expected value for the recursive linear search.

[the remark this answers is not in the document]
ZoomRmc: Yep, that's why I added the clarification, since different direction means different results (first first / last first).

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+      return some(Natural(i))
+  none(Natural) # `key` not found
+
+func recursiveLinearSearch*[T](arr: openArray[T], key: T, idx: Nat = arr.low.Nat): OptNat=
+  # Recursion is another method for linear search
+  # we can just replace the for loop with recursion.
+
+  # `none(Natural)` is returned when the array is traversed completely
+  # and no key is matched, or when `arr` is empty and has a length of 0
+  if idx >= arr.high:
```
dlesnoff, line 47:
```suggestion
  if idx > arr.high:
```
Add a test with a valid key at the end of the array.

searches/linear_search.nim
```diff
@@ -0,0 +1,85 @@
...
+    return none(Natural)
+  if arr[idx] == key:
+    return some(idx)
+  recursiveLinearSearch(arr, key, idx + 1)
+
+
+when isMainModule:
+  import unittest
+
+  template checkLinearSearch[T](arr: openArray[T], key: T, expectedIdx: OptNat =
```
dlesnoff, line 57:
```suggestion
  template checkLinearSearch[T](arr: openArray[T], key: T, expectedIdx: OptNat): untyped =
```

dlesnoff (changes requested):
You are almost there.

SatinWuker:
Ahhh finally

SatinWuker (commented):
I have double-checked, and the code works 👍

searches/linear_search.nim
```diff
@@ -0,0 +1,80 @@
...
+## or in the opposite direction for the recursive version, as implemented in this module.
+## The search is finished and terminated once the target element is located.
+## https://www.simplilearn.com/tutorials/data-structure-tutorial/linear-search-algorithm
+##
+## Time Complexity: O(N)
+## Space Complexity in for-loop linear search: O(1)
+## Space Complexity in recursive linear search: O(n)
+## This is because of the call stack operation,
+## and there is no CPS automatic optimization in Nim
+## CPS: https://en.wikipedia.org/wiki/Continuation-passing_style
```
dlesnoff, line 16:
```suggestion
```

searches/linear_search.nim
```diff
@@ -0,0 +1,80 @@
...
+
+
+import std/options
+
+type
+  Nat = Natural
+  OptNat = Option[Natural]
+
+func linearSearch*[T](arr: openArray[T], key: T): OptNat = 
+  # key is the value for matching in the array
```
dlesnoff, line 35:
```suggestion
  # key is the value we are searching for in the array
```

searches/linear_search.nim
```diff
@@ -0,0 +1,80 @@
...
+    if val == key:
+      return some(Natural(i))
+  none(Natural) # `key` not found
+
+func recursiveLinearSearch*[T](arr: openArray[T], key: T, idx: Nat = arr.low.Nat): OptNat=
+  # Recursion is another method for linear search
+  # we can just replace the for loop with recursion.
+
+  # `none(Natural)` is returned when the array is traversed completely
+  # and no key is matched, or when `arr` is empty and has a length of 0
```
dlesnoff, line 46:
```suggestion
  # Recursion is another method for linear search.
  # Recursive calls replace the for loop.

  # `none(Natural)` is returned when the array is traversed completely
  # and no key is matched, or when `arr` is empty.
```

searches/linear_search.nim
```diff
@@ -0,0 +1,80 @@
...
+  suite "Linear search":
+    test "Search in an empty array":
+      var arr: array[0, int]
+      checkLinearSearch(arr, 5, none(Natural))
+
+    test "Search in an int array matching with a valid value":
+      var arr = [0, 3, 1, 4, 5, 6]
+      checkLinearSearch(arr, 5, some(Natural(4)))
+
+    test "Search in an int array matching with an invalid value":
```
dlesnoff, line 70:
```suggestion
    test "Search in an int array for a missing value":
```

dlesnoff (changes requested):
The code looks much better to me. I am about to merge this, but the comments do not reflect the last code changes. The writing style can be improved.

searches/linear_search.nim
dlesnoff: Test?

dlesnoff:
Please run `nim r .scripts/directory.nim > DIRECTORY.md` and add the file to your PR.

searches/linear_search.nim
```diff
@@ -0,0 +1,80 @@
+## Linear Search
+## =============
+## Linear search is the simplest but least efficient algorithm
+## to search for an element in a data set.
+## It examines each element until it finds a match,
+## starting at the beginning of the data set for the iterative version,
+## or in the opposite direction for the recursive version, as implemented in this module.
```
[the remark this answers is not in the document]
SatinWuker: Excuse me, what do you mean by that

searches/linear_search.nim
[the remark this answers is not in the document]
SatinWuker: Thank you for your research and detailed explanation! I included it in the comments.

searches/linear_search.nim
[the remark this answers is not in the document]
dlesnoff: Both recursive and iterative versions start at the beginning of the array.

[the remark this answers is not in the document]
SatinWuker: ok, I get it.

searches/linear_search.nim
```diff
@@ -0,0 +1,88 @@
+## Linear Search
+## =============
+## Linear search is the simplest but least efficient searching algorithm
+## to search for an element in a data set.
+## It examines each element until it finds a match,
+## starting at the beginning of the data set toward the end.
+## The search ends when the element is located or when the end of the array is reached.
+## https://en.wikipedia.org/wiki/Linear_search
+##
+## Time Complexity: O(n)
```
dlesnoff, line 10:
```suggestion
## Time Complexity: O(n) where n is the length of the array.
```

searches/linear_search.nim
```diff
@@ -0,0 +1,88 @@
+## Linear Search
+## =============
+## Linear search is the simplest but least efficient searching algorithm
+## to search for an element in a data set.
```
dlesnoff, line 4: data set being an array here. The algorithm could apply to a linked list too, or anything that could be sequentially indexed, well anything that can be stored in an array.
```suggestion
## to search for an element in an array.
```

SatinWuker:
> Please run `nim r .scripts/directory.nim > DIRECTORY.md` and add the file to your PR.

DIRECTORY.md updated 👌

dlesnoff (approved):
You nailed it. Solid first contribution that will serve as reference for others.

SatinWuker:
> You nailed it. Solid first contribution that will serve as reference for others.

Thank you so much for helping me from the beginning all the way to the end! This process is just like a linear search hahaha.
I really learned a lot from you guys, and I will write better-quality code and maintain this project.

ZoomRmc:
Thank you for the contribution and sorry for a bit laborious review process.

Nim has an exceptionally high educative potential due to its uncluttered and intuitive syntax, so adhering to strict guidelines within this repository right from the get go can showcase it in a good light and stand out in comparison with implementations in other languages.

PR #33, opened 2023-06-20
Fix linear search

### LINKED ISSUE #31: CI failure: linear search algorithm, string array and recursive algorithm.
Currently, CI fails due to the linear search algorithm.
There is an issue with the recursive algorithm for string arrays.

```nim
proc recursiveLinearSearch*[T](arr: openArray[T], key: T, idx: Nat = arr.low.Nat): OptNat=
  # Recursion is another method for linear search.
  # Recursive calls replace the for loop.

  # `none(Natural)` is returned when the array is traversed completely
  # and no key is matched, or when `arr` is empty.
  if idx > arr.high:
    return none(Natural)
  if arr[idx] == key:
    return some(idx)
  echo idx, " ", arr[idx], " ", key
  recursiveLinearSearch(arr, key, idx + 1)
```
for the test:
```nim
var arr = ["0", "c", "a", "u", "5", "7"]
check recursiveLinearSearch(arr, "5") == some(Natural(4))
```
outputs:
```raw
0  5
1  5
2  5
3  5
4  5
5  5
```
Expected output:
```raw
0 0 5
1 c 5
2 a 5
3 u 5
```
We could maybe fix it by converting the example test array to a sequence:
```raw
var arr = @["0", "c", "a", "u", "5", "7"]
```

dlesnoff:
Fix #31 and CI. Ran nimpretty directly.

### DIFF
--- a/searches/linear_search.nim
+++ b/searches/linear_search.nim
@@ -33,7 +33,10 @@ runnableExamples:
 import std/options
 
 func linearSearch*[T](arr: openArray[T], key: T): Option[Natural] =
-  # key is the value we are searching for in the array.
+  ## Searches for the `key` in the array `arr` and returns its absolute index (counting from 0)
+  ## in the array.
+  ## .. Note:: For arrays indexed with a range type or an enum the returned value
+  ##  may not be consistent with the indexing of the initial array.
   for i, val in arr.pairs():
     if val == key:
       return some(Natural(i))
@@ -41,8 +44,11 @@ func linearSearch*[T](arr: openArray[T], key: T): Option[Natural] =
 
 func recursiveLinearSearch*[T](
   arr: openArray[T], key: T, idx: Natural = Natural(0)): Option[Natural] =
-  # Recursion is another method for linear search.
-  # Recursive calls replace the for loop.
+  ## Searches for the `key` in `arr` and returns its absolute index (counting from 0)
+  ## in the array. Search is performed in a recursive manner.
+  ## .. Note:: For arrays indexed with a range type or an enum the returned value
+  ##  is not consistent with the indexing of the parameter array if `arr.low` is not 0.
+  # Recursive calls replace the `for` loop in `linearSearch`.
 
   # `none(Natural)` is returned when the array is traversed completely
   # and no key is matched, or when `arr` is empty.
--- a/searches/linear_search.nim
+++ b/searches/linear_search.nim
@@ -35,14 +35,14 @@ type
   Nat = Natural
   OptNat = Option[Natural]
 
-func linearSearch*[T](arr: openArray[T], key: T): OptNat = 
+func linearSearch*[T](arr: openArray[T], key: T): OptNat =
   # key is the value we are searching for in the array.
   for i, val in arr.pairs():
     if val == key:
       return some(Natural(i))
   none(Natural) # `key` not found
 
-func recursiveLinearSearch*[T](arr: openArray[T], key: T, idx: Nat = arr.low.Nat): OptNat=
+proc recursiveLinearSearch*[T](arr: openArray[T], key: T, idx: Nat = arr.low.Nat): OptNat=
   # Recursion is another method for linear search.
   # Recursive calls replace the for loop.
 
@@ -80,7 +80,7 @@ when isMainModule:
       checkLinearSearch(arr, '5', some(Natural(4)))
 
     test "Search in a string array matching with a string matching value":
-      var arr = ["0", "c", "a", "u", "5", "7"]
+      var arr = @["0", "c", "a", "u", "5", "7"]
       checkLinearSearch(arr, "5", some(Natural(4)))
 
     test "Search in an int array with a valid key at the end":
--- a/searches/linear_search.nim
+++ b/searches/linear_search.nim
@@ -16,6 +16,8 @@
 ## Therefore the recursive linear search is less efficient than the for-loop-based one.
 
 runnableExamples:
+  import std/options
+
   var arr1 = [0, 3, 1, 4, 5, 6]
   doAssert linearSearch(arr1, 5) == some(Natural(4))
   doAssert recursiveLinearSearch(arr1, 5) == some(Natural(4))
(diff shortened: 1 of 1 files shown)


ZoomRmc (commented):
@beef331 is right. Since the logic has already been changed to iterate the array in ascending manner, better to change the default to `Natural(0)`.

I would also removed the custom type aliases, it trades better locality for shorter lines (that still wrap), so no win as I see it.

What's the reason for changing the `func` to `proc`?

ZoomRmc (approved):

searches/linear_search.nim
```diff
@@ -80,7 +79,7 @@ when isMainModule:
       checkLinearSearch(arr, '5', some(Natural(4)))
 
     test "Search in a string array matching with a string matching value":
```
SatinWuker, line 81: Since you added `@` on line 82, I would do 
```suggestion
    test "Search in a string sequence matching with a string matching value":
```

SatinWuker (changes requested):
Maybe just do this, to avoid confusion. Overall nim prettify has done illustriously 👍 
And indeed, adding type definition to `Natural` and `Option[Natural]` looks redundant and verbose. well done!

pietroppeter:
As a note on linearSearch I think current version is buggy in case of arrays with negative indices (which are allowed for open array), see https://play.nim-lang.org/#ix=4yJs

```nim
import std/options

func linearSearch*[T](arr: openArray[T], key: T): Option[Natural] =
  # key is the value we are searching for in the array.
  for i, val in arr.pairs():
    if val == key:
      return some(Natural(i))
  none(Natural) # `key` not found

type
  NegIdxArray = array[-1 .. 1, int]

let
  a: NegIdxArray = [1,2,3]

echo linearSearch(a, 1) # some(0)
```

Correct answer should be `some(-1)`

dlesnoff:
@pietroppeter This is also the case for any array with a custom starting index, isn't it?

beef331:
This is the case for any array that does not start at 0 or use a data type that starts as 0, it's probably best to just have an `openArray` and a `array[Idx, T]` search.

dlesnoff:
Having two functions for each would be contrary to the DRY - Don't Repeat Yourself principle. That feels like unnecessary duplicate code for me.

beef331:
Sure but they have to be  specialized. `array[bool, T]` or `array[MyEnum, T]` gets `Option[Natural]` which requires a type conversion back to index type by the user. The most reusable that should work for all types and the least amount of DRY breaking is as follows: 

```nim

func linearSearchImpl[T](arr: openArray[T], key: T): int =
  for i, val in arr.pairs():
    if val == key:
      return i
  -1
  
func linearSearch*[T](arr: openArray[T], key: T): Option[Natural] =
  let val = linearSearchImpl(arr, key)
  if val >= 0:
    some(Natural(val))
  else:
    none(Natural)
    
func linearSearch*[Idx, T](arr: array[Idx, T], key: T): Option[Idx] =
  let val = linearSearchImpl(arr, key)
  if val >= 0:
    some(Idx(ord(arr.low) + val))
  else:
    none(Idx)  
```

pietroppeter:
Ah, I did not know that openarray did not support low and high. checked here indeed it has 0 as low always for openarray: https://play.nim-lang.org/#ix=4yJD

beef331:
Well they support it as you've shown, but `openArray` is explicitly a type erasure so yea it loses the source type information. It's implemented as a `ptr T, len` afterall.

dlesnoff:
@beef331 I do not want to merge your modifications. It adds much complexity to handle an often superfluous feature of array reindexing. I prefer to keep the implementation as simple as possible.

While one of our implementation's goals is « easy integration in one's code project », our primary goal is pedagogical algorithm teaching. I will use your code if indexed arrays reveal to be needed in the future.

ZoomRmc:
> @beef331 I do not want to merge your modifications. It adds much complexity to handle an often superfluous feature of array reindexing. I prefer to keep the implementation as simple as possible.
> 
In this case it would be nice to document that the function does return the absolute index and not a corresponding index value for range/enum-indexed arrays.

searches/linear_search.nim
```diff
@@ -28,21 +30,17 @@ runnableExamples:
...
-
 import std/options
 
-type
-  Nat = Natural
-  OptNat = Option[Natural]
-
-func linearSearch*[T](arr: openArray[T], key: T): OptNat = 
+func linearSearch*[T](arr: openArray[T], key: T): Option[Natural] =
   # key is the value we are searching for in the array.
```
ZoomRmc, line 36:
```suggestion
  ## Searches for the `key` in `arr` and returns its absolute index (counting from 0)
  ## in the array.
  ## .. Note:: For arrays indexed with a range type or an enum the returned value
  ##  may not be consistent with the indexing of the initial array.
```

searches/linear_search.nim
```diff
@@ -28,21 +30,17 @@ runnableExamples:
...
   for i, val in arr.pairs():
     if val == key:
       return some(Natural(i))
   none(Natural) # `key` not found
 
-func recursiveLinearSearch*[T](arr: openArray[T], key: T, idx: Nat = arr.low.Nat): OptNat=
+func recursiveLinearSearch*[T](
+  arr: openArray[T], key: T, idx: Natural = Natural(0)): Option[Natural] =
   # Recursion is another method for linear search.
   # Recursive calls replace the for loop.
```
ZoomRmc, line 45:
```suggestion
  ## Searches for the `key` in `arr` and returns its absolute index (counting from 0)
  ## in the array. Search is performed in a recursive manner.
  ## .. Note:: For arrays indexed with a range type or an enum the returned value
  ##  may not be consistent with the indexing of the initial array.
  # Recursive calls replace the `for` loop in `linearSearch`.
```

ZoomRmc (changes requested):

searches/linear_search.nim
dlesnoff: Thanks for the comments suggestions.

Panquesito7 (approved):
Approving to allow the merge. Thanks. 👍

PR #35, opened 2023-06-21
Implement Gale-Shapley stable matching

ZoomRmc:
This is an adaptation of my much [earlier implementation](https://rosettacode.org/wiki/Stable_marriage_problem#Nim) made for the RosettaCode project. The additional requirements of the task were moved to the runnableExamples block.

The code was updated, thoroughly checked and commented.

The name for the category is similar to one used for earlier [C#](https://github.com/TheAlgorithms/C-Sharp/tree/master/Algorithms/Problems/StableMarriage) and [Lua](https://github.com/TheAlgorithms/Lua/tree/main/src/misc/stable_marriage) implementations.

### DIFF
--- a/stable_marriage/gale_shapley.nim
+++ b/stable_marriage/gale_shapley.nim
@@ -1,0 +1,267 @@
+## Gale-Shapley
+## ============
+##
+## This module implements the classical Gale-Shapley_ algorithm for solving
+## the stable matching problem.
+##
+## The following example uses the implemented algorithm to solve the task of
+## stable matching as posed by RosettaCode_:
+##
+## .. _Gale-Shapley: https://en.wikipedia.org/wiki/Gale%E2%80%93Shapley_algorithm
+## .. _RosettaCode:  https://rosettacode.org/wiki/Stable_marriage_problem
+##
+## * Finds a stable set of matches ("engagements").
+## * Randomly destabilizes the matches and checks the results for stability.
+##
+## Example starts with a discarded string containing a possible output:
+runnableExamples:
+  discard """
+  abe 💑 ivy, bob 💑 cath, col 💑 dee, dan 💑 fay, ed 💑 jan,
+  fred 💑 bea, gav 💑 gay, hal 💑 eve, ian 💑 hope, jon 💑 abi
+  Current matching stability: ✓ Stable
+  Swapping matches for random contenders: bob <=> gav
+  abe 💑 ivy, bob 💑 gay, col 💑 dee, dan 💑 fay, ed 💑 jan,
+  fred 💑 bea, gav 💑 cath, hal 💑 eve, ian 💑 hope, jon 💑 abi
+  Current matching stability: ✗ Unstable
+  💔 bob prefers cath over gay
+  💔 cath prefers bob over gav
+  """
+
+  import std/[random, strutils, sequtils, strformat, options]
+
+  const
+    MNames = ["abe", "bob", "col", "dan", "ed", "fred", "gav", "hal", "ian", "jon"]
+    FNames = ["abi", "bea", "cath", "dee", "eve", "fay", "gay", "hope", "ivy", "jan"]
+    MPreferences = [
+      ["abi", "eve", "cath", "ivy", "jan", "dee", "fay", "bea", "hope", "gay"],
+      ["cath", "hope", "abi", "dee", "eve", "fay", "bea", "jan", "ivy", "gay"],
+      ["hope", "eve", "abi", "dee", "bea", "fay", "ivy", "gay", "cath", "jan"],
+      ["ivy", "fay", "dee", "gay", "hope", "eve", "jan", "bea", "cath", "abi"],
+      ["jan", "dee", "bea", "cath", "fay", "eve", "abi", "ivy", "hope", "gay"],
+      ["bea", "abi", "dee", "gay", "eve", "ivy", "cath", "jan", "hope", "fay"],
+      ["gay", "eve", "ivy", "bea", "cath", "abi", "dee", "hope", "jan", "fay"],
+      ["abi", "eve", "hope", "fay", "ivy", "cath", "jan", "bea", "gay", "dee"],
+      ["hope", "cath", "dee", "gay", "bea", "abi", "fay", "ivy", "jan", "eve"],
+      ["abi", "fay", "jan", "gay", "eve", "bea", "dee", "cath", "ivy", "hope"]
+    ]
+    FPreferences = [
+      ["bob", "fred", "jon", "gav", "ian", "abe", "dan", "ed", "col", "hal"],
+      ["bob", "abe", "col", "fred", "gav", "dan", "ian", "ed", "jon", "hal"],
+      ["fred", "bob", "ed", "gav", "hal", "col", "ian", "abe", "dan", "jon"],
+      ["fred", "jon", "col", "abe", "ian", "hal", "gav", "dan", "bob", "ed"],
+      ["jon", "hal", "fred", "dan", "abe", "gav", "col", "ed", "ian", "bob"],
+      ["bob", "abe", "ed", "ian", "jon", "dan", "fred", "gav", "col", "hal"],
+      ["jon", "gav", "hal", "fred", "bob", "abe", "col", "ed", "dan", "ian"],
+      ["gav", "jon", "bob", "abe", "ian", "dan", "hal", "ed", "col", "fred"],
+      ["ian", "col", "hal", "gav", "fred", "bob", "abe", "ed", "jon", "dan"],
+      ["ed", "hal", "gav", "abe", "bob", "jon", "col", "ian", "fred", "dan"]
+    ]
+    # Recipient ids in descending order of preference
+    ContenderPrefs = initContenderPrefs(MPreferences, FNames)
+    # Preference score for each contender's id
+    RecipientPrefs = initRecipientPrefs(FPreferences, MNames)
+
+  proc randomPair(max: Positive): (Natural, Natural) =
+    ## Returns a random fair pair of non-equal numbers up to and including `max`
+    let a = rand(max)
+    var b = rand(max - 1)
+    if b == a: b = max
+    (a.Natural, b.Natural)
+
+  proc perturbPairs(ms: var Matches) =
+    randomize()
+    let (a, b) = randomPair(ms.contenderMatches.len - 1)
+    echo "Swapping matches between random contenders: ",
+          MNames[a], " <=> ", MNames[b]
+    template swap(arr: var openArray[int]; a, b: int) = swap(arr[a], arr[b])
+    ms.contenderMatches.swap(a, b)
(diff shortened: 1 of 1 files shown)


stable_marriage/gale_shapley.nim
```diff
@@ -0,0 +1,274 @@
...
+    # Queue holding the currently considered "preference score"
+    # (idx of contenderPrefs) for each contender.
+    contQueue = newSeqWith(rosterLen, 0)
+  template match(c, r: Natural) =
+    # Recipient accepts contender, the match is stored
+    contMatches[c] = r
+    recMatches[r] = c
+  while contMatches.contains(-1): # While exist unmatched contenders...
+    for c in 0..<rosterLen: # for each contender...
+      if contMatches[c] == -1: # if they are yet unmatched...
```
ZoomRmc, line 199: So I thought for a bit if I can make it work without this scan-through at the beginning of each big cycle followed by a filter and came up with the following code.

It leverages an additional sequence used as a dequeue for keeping track of unmatched contenders. This allows to drop the `contains` and `if ... == -1` parts, but the trade-off is that the order of traversal changes. Since `del` moves the last element at the beginning of the sequence, this implementation often oscillates between two contenders and scan goes down their corresponding preferences until a match is found for either of them. I suspected this imbalance could increase the number of total iterations until a stable matching for all participants is found.

To test it I pre-generated a random input table for 2048 contenders+2048 recipients and measured the performance of both versions. Surprisingly, the number of iterations turned out to be equal (14874). The difference in execution time is negligible and may even be worse for the new version, and the code is a bit more complex. So, not an improvement overall, so I leave it here just to document that it's been tried.

```nim
func stableMatchingNew*(contenderPrefs, recipientPrefs: openArray[
    Ranking]): Matches =
  let rosterLen = recipientPrefs.len
  var
    # Initializing result sequences with -1, meaning "unmatched"
    recMatches = newSeqWith(rosterLen, -1)
    contMatches = newSeqWith(rosterLen, -1)
    # Queue holding the currently considered "preference score" (idx of
    # contenderPrefs) for each contender, initialized with the most preferred
    contQueue = newSeqWith(rosterLen, 0)
    unmatchedContenders = toSeq(0..<rosterLen) # Queue of vacant contenders
    # A seq is used as a dequeue, traversal of unmatched contenders in unordered
  template match(c, r: Natural) =
    # Recipient accepts contender, the match is stored
    contMatches[c] = r
    recMatches[r] = c
  while unmatchedContenders.len > 0: # While exist unmatched contenders...
    let c = unmatchedContenders[0]
    unmatchedContenders.del(0) # Remove the contender from the vacant list
    let r = contenderPrefs[c][contQueue[c]] # Proposing to the queued recipient
    inc(contQueue[c]) # Queue next preferred recipient for this contender
    assert contQueue[c] <= rosterLen
    let rivalMatch = recMatches[r] # Recipient's match index or -1 (vacant)
    if rivalMatch == -1: # Recipient is also unmatched
      match(c, r)
    # Recipient is matched, but contender is more preferable
    elif recipientPrefs[r][c] < recipientPrefs[r][rivalMatch]:
      contMatches[rivalMatch] = -1 # Vacate current recipient's match and...
      unmatchedContenders.add(rivalMatch) # add them to the list
      match(c, r)
    else: # Contender's proposition is rejected
      unmatchedContenders.add(c)
  Matches(contenderMatches: contMatches, recipientMatches: recMatches)
```

PS: I wonder if there's a difference if tested on some data bearing closer resemblance to the real world, with some participants being evidently more popular and some less so.

dlesnoff: I promise to have a look at it this weekend. It seems like you hit a strange interesting corner case with this optimisation attempt.
I do not have books as references for the stable marriage problem.
I know some «real-world» applications:
– advertisement for users with preferences in e.g. search engines. (There are maybe better formulations of this problem than stable marriage).
– cryptography, but don't remember exactly where.

dlesnoff: You need a sorted list of indexes of unmatched contenders.
Since you need the minimal index, a heap would be best suited for this.

stable_marriage/gale_shapley.nim
```diff
@@ -0,0 +1,274 @@
...
+      if recipientPrefs[checkedRec][checkedRival] > recipientPrefs[checkedRec][c]:
+        let clashC = (id: c.Natural, match: checkedRec.Natural,
+                      prefers: curMatch.Natural)
+        let clashR = (id: checkedRec.Natural, match: c.Natural,
+                      prefers: checkedRival.Natural)
+        return some((clashC, clashR))
+  none((Clash, Clash))
+
+
+when isMainModule:
```
dlesnoff, line 238: I would add some simple edge cases there:
- I actually think that algorithms have been designed for cases when contenders.len < recipients.len or vice-versa.
It might be interesting to add one example to check an assertion has been raised.

stable_marriage/gale_shapley.nim
```diff
@@ -0,0 +1,274 @@
...
+{.push raises: [].}
+
+import std/options
+from std/sequtils import newSeqWith
+
+type
+  Ranking*[T: int] = concept c ## A wide concept allowing `stableMatching`
+                    ## to accept both arrays and sequences in an openArray.
+    c.len is Ordinal
+    c[int] is T
```
dlesnoff, line 132: Are `concept`s not experimental anymore?

dlesnoff (commented):
Not a great review, but I think that the implementation is quite fine!
I will make another pass in a few days.

stable_marriage/gale_shapley.nim
```diff
@@ -0,0 +1,274 @@
...
+    stdout.write "Current matching stability: "
+    let areStable = checkPairStability(matches, RecipientPrefs, ContenderPrefs)
+    doAssert areStable
+
+  perturbPairs(matches)
+  block:
+    echo render(matches.contenderMatches, MNames, FNames)
+    stdout.write "Current matching stability: "
+    let areStable = checkPairStability(matches, RecipientPrefs, ContenderPrefs)
+    doAssert areStable == false
```
dlesnoff, line 120:
```suggestion
    doAssert !areStable
```
Less readable, but I add it as a consideration to make.

dlesnoff, line 120: These two blocks appear twice. Template?
```suggestion
  template outputMatchesWithStabilityStatus(status: bool) =
    block:
      echo render(matches.contenderMatches, MNames, FNames)
      stdout.write "Current matching stability: "
      let areStable = checkPairStability(matches, RecipientPrefs, ContenderPrefs)
      doAssert areStable == status
  
  outputMatchesWithStabilityStatus(true)
  perturbPairs(matches)
  outputMatchesWithStabilityStatus(false)
```

stable_marriage/gale_shapley.nim
ZoomRmc: They are, but the old ones have been available and working for a very long time. I don't think there's anything wrong with using them, as I don't think there's a risk of them being dropped completely, only replaced by a new style concepts that we can adapt the code to as necessary.

stable_marriage/gale_shapley.nim
```diff
@@ -0,0 +1,274 @@
...
+    randomize()
+    let (a, b) = randomPair(ms.contenderMatches.len - 1)
+    echo "Swapping matches between random contenders: ",
+          MNames[a], " <=> ", MNames[b]
+    template swap(arr: var openArray[int]; a, b: int) = swap(arr[a], arr[b])
+    ms.contenderMatches.swap(a, b)
+    ms.recipientMatches.swap(ms.contenderMatches[a], ms.contenderMatches[b])
+
+  func str(c: Clash; aNames, bNames: openArray[string]): string =
+    &"\u{0001F494} {aNames[c.id]} prefers {bNames[c.prefers]} over {bNames[c.match]}"
```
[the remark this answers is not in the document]
dlesnoff: Either we wrap a type for Clash or aNames and bNames are global variables. I want neither. I resolve the conv'.

stable_marriage/gale_shapley.nim
```diff
@@ -0,0 +1,274 @@
+## Gale-Shapley
+## ============
```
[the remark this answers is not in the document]
ZoomRmc: Perhaps to avoid having it as a hard requirement we could open a long-standing issue for tracking proper doc generation for merged modules.

stable_marriage/gale_shapley.nim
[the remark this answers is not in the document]
ZoomRmc: My personal style usually permits a bit more wordiness than a median (at list for Nim), I think. I'm also wary of overloads unless they work absolutely smoothly, and this is not that case.

stable_marriage/gale_shapley.nim
ZoomRmc: BTW, since this repo is educational we're much more free to use and showcase all the cool features of the language, compared to production situation.Since Nim doesn't have interfaces/traits, I think it's important to show the next best thing available.

stable_marriage/gale_shapley.nim
ZoomRmc: Is having a *sorted* list a requirement? I haven't thought much about it, but looks like the order doesn't matter much.

Also, my hunch is an overhead of a heap is not worth it for this algorithm. Scanning an array or ints is very cheap, and I think there's a merit in an implementation that doesn't use anything besides arrays/seqs and is still efficient.

stable_marriage/gale_shapley.nim
```diff
@@ -0,0 +1,274 @@
...
+##
+## This module implements the classical Gale-Shapley_ algorithm for solving
+## the stable matching problem.
```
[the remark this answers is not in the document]
ZoomRmc: Great idea! I think it's better to add it in a separate PR. I'll make a stub later.

Definitely not Streaming, and I don't think Online either, as GS guarantees the best solution for _all_ receivers, everything will change when the lists are changed.

stable_marriage/gale_shapley.nim
ZoomRmc: Huh, turns out, the implementation is structured in a way where it's not so easy to pass invalid data in. Any suggestions?

dlesnoff: `doAssertRaises( (name of the defect))`
I can't remember the exact syntax since it does not appear in the docs of std/unittest anymore.

dlesnoff: Sorry, I was unclear. I wanted that you keep your implementation with the assumption that the sets are of the same length and just add an example with a doAssertRaises, to tell that your implementation catches erroneous inputs.
I think this is actually enough and will accept it like this.

stable_marriage/gale_shapley.nim
dlesnoff: Sure, it is an overhead for this algorithm. Well, you are right, the order should not matter.

dlesnoff (approved):
Great contrib' !

PR #41, opened 2023-06-23
chore: run `apt-get update` before running other commands

Panquesito7:
Things added/changed:

- Run `sudo apt-get update` before running other commands on the Gitpod Dockerfile.

### DIFF
--- a/.gitpod.dockerfile
+++ b/.gitpod.dockerfile
@@ -1,6 +1,6 @@
 FROM gitpod/workspace-base
 
-RUN curl https://nim-lang.org/choosenim/init.sh -sSf -o install_nim.sh \
+RUN sudo apt-get update && curl https://nim-lang.org/choosenim/init.sh -sSf -o install_nim.sh \
     && chmod +x ./install_nim.sh \
     && ./install_nim.sh -y \
     && rm install_nim.sh \


dlesnoff:
Isn't an `apt-get upgrade` necessary after an update?

Panquesito7:
> Isn't an `apt-get upgrade` necessary after an update?

Sometimes, not all of the packages are updated, even on the first start, AFAIK.

dlesnoff (approved):
Looks good to me.

vil02:
@Panquesito7 I know that this was already merged, but I would like to know why do you think this change is needed? Is it recommended to run `sudo apt-get update` before executing the `https://nim-lang.org/choosenim/init.sh` script?

dlesnoff:
`curl` might not be up to date.

vil02:
`sudo apt-get update` only updates the _package index_. In this case `sudo apt-get upgrade` is needed.

A side remark: in my opinion having a not updated `curl' is a smaller problem than using `sudo` in a docker file.

dlesnoff:
I will check Docker's security recommendations tomorrow.

vil02:
After thinking about it, I have a strong opinion: this change should be reverted. The reasons are as follows:
- `apt-get update` updates the _index_ (and this is not needed in this case),
- adding `upgrade` in order to update installed packages like `curl` does not make sense: this should be done _inside_ the image `gitpod/workspace-base`.

@dlesnoff @Panquesito7 if you agree with me, I will make a suitable PR.

dlesnoff:
I revert it, and I wait for a better suggestion.

PR #47, opened 2023-06-27
chore: `gc`->`mm`, move back to `config.nims`, fixes #26

### LINKED ISSUE #26: chore: move compiler defaults from `nim.cfg` to `config.nims`
At this moment, in some cases setting the default `gc`/`mm` setting for the compiler results in [a compilation error](https://github.com/TheAlgorithms/Nim/pull/21#issuecomment-1574355105).
The workaround for this is moving the setting to the `nim.cfg` file. Having an additional configuration file is not great, so as soon as the fix for the bug lands in stable the setting should be moved back to the `config.nims` and `nim.cfg` should be removed.

This is a PR that sould have fixed the issue: https://github.com/nim-lang/Nim/pull/19455. Not sure why we're still hitting it.

ZoomRmc:
Fixes #26 and removes the now unused `nim.cfg`

Setting the `mm`/`gc` switch turned out to be position-sensitive. Moving it below the imports helps. Will not be necessary as soon as the fixes from devel land in the stable compiler branch. The switch is renamed to the not deprecated `mm` version.

This also silences the `BareExcept` warning when running tests with `nim test` that mostly stem from the `std/unittest`. The used compiler switch (`warning[BareExcept]:off`) surprisingly works fine on stable but didn't work reliably on devel (should be fixed with nim-lang/Nim/pull/21390 by now).

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -13,6 +13,8 @@ import std/[os, sequtils]
 from std/strutils import startsWith, endsWith
 from std/strformat import `&`
 
+--mm:arc # TODO: check if still needs to be below imports on Nim > 1.6.12
+
 const IgnorePathPrefixes = ["."]
 
 func isIgnored(path: string): bool =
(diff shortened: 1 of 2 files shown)


dlesnoff (approved):
Good news!
Thanks for the fix.
I am glad to get rid of these warnings 👍🏽

PR #51, opened 2023-06-30
Added Stable Marriage/Matching overview; more tests

ZoomRmc:
* Added an overview to the "Stable Marriage" category, as discussed in https://github.com/TheAlgorithms/Nim/pull/35#discussion_r1246584706
* Added more docs and a test case to `gale_shapley.nim`
* Modified the `.scripts/directory.nim` to link to category readmes, if found.
* Updated the `DIRECTORY.md`

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -17,6 +17,9 @@
   * [Binary Search](searches/binary_search.nim)
   * [Linear Search](searches/linear_search.nim)
 
+## [Stable Marriage](stable_marriage/README.md)
+  * [Gale-Shapley](stable_marriage/gale_shapley.nim)
+
 ## Strings
   * [Check Anagram](strings/check_anagram.nim)
 
--- a/stable_marriage/README.md
+++ b/stable_marriage/README.md
@@ -6,7 +6,7 @@ The Stable Marriage Problem[^gs] involves finding a pairing between two equally 
 
 Stable Matching has real-world applications in various domains such as assigning graduating medical students to hospitals, assigning users to servers in distributed Internet services, and market design.
 
-The theory of stable allocations and market design, including the Stable Matching problem, was recognised with the Nobel Prize in Economic Sciences in 2012.
+The theory of stable allocations and the practice of market design, which includes the Stable Matching problem, was recognised with the Nobel Prize in Economic Sciences in 2012, awarded to Alvin E. Roth and Lloyd S. Shapley.
 
 ## The Gale-Shapley Algorithm
 The [Gale-Shapley algorithm](gale_shapley.nim), proposed by David Gale and Lloyd Shapley, is a widely used solution for the Stable Marriage Problem. It guarantees a stable matching (marriage) for all participants and has a time complexity of `O(n^2)`, where `n` is the number of contenders or receivers (men or women).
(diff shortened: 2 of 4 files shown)


DIRECTORY.md
```diff
@@ -17,6 +17,9 @@
   * [Binary Search](searches/binary_search.nim)
   * [Linear Search](searches/linear_search.nim)
 
+## [Stable Marriage](stable_marriage/README.md)
+  * [Gale-Shapley](stable_marriage/gale_shapley.nim)
+
```
dlesnoff, line 22: Well done! I hadn't thought this needed a change in the `directory.nim` script!

stable_marriage/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Other algorithms
+
+While the Gale-Shapley algorithm is popular due to its efficiency and stability guarantees, other algorithms[^other] exist with different trade-offs in stability, optimality, and computational complexity.
+
+### Vande Vate algorithm
+
+Vande Vate's algorithm[^rvv], also known as the Roth and Vande Vate (RVV) mechanism, is an alternative algorithm with the following differences, compared to Gale-Shapley:
+
+- Starts with an arbitrary initial assignment of contenders to receivers, not with an empty matching.
+- Allows introducing participants incrementally and let them iteratively reach a stable matching.
```
dlesnoff, line 31: I think that this is due to the `online` nature of the Algorithm and we could introduce the concept here.

stable_marriage/README.md
```diff
@@ -0,0 +1,39 @@
...
+### Vande Vate algorithm
+
+Vande Vate's algorithm[^rvv], also known as the Roth and Vande Vate (RVV) mechanism, is an alternative algorithm with the following differences, compared to Gale-Shapley:
+
+- Starts with an arbitrary initial assignment of contenders to receivers, not with an empty matching.
+- Allows introducing participants incrementally and let them iteratively reach a stable matching. 
+- Accommodates scenarios with changing preferences.
+
+## Related Problems:
+Several related problems stem from the Stable Matching Problem, including stable matching with indifference, the stable roommates problem, the hospitals/residents problem, and matching with contracts. These variations address scenarios with slightly different sets of constraints, such as: ties in preferences, a single pool of participants, multiple residents and hospitals/colleges, and different contract terms.
```
dlesnoff, line 35: Can these problems be added to this repository too?
I think we should envision the project as potentially wider and present not only stable matching but matching problems in general.

dlesnoff (changes requested):
Thanks for the script changes, and the extra test!

The README is maybe too focused on the stable marriage problem which would then only accept two or three algorithms. We should put the stable roommate's problem directly in this repository, and find a common name to group them.

stable_marriage/README.md
```diff
@@ -0,0 +1,39 @@
+# Stable Marriage Problem
+> Also known as the Stable Matching Problem
+
+## Problem overview
+The Stable Marriage Problem[^gs] involves finding a pairing between two equally sized sets of elements based on their preferences for each other. In the context of the Stable Marriage Problem, there are `n` men (contenders) and `n` women (receivers) with ranked preferences for members of the opposite group. The objective is to match men and women together so that there are no pairs in which **both** participants would prefer some other pairing over their current partners.
+
+Stable Matching has real-world applications in various domains such as assigning graduating medical students to hospitals, assigning users to servers in distributed Internet services, and market design.
+
+The theory of stable allocations and market design, including the Stable Matching problem, was recognised with the Nobel Prize in Economic Sciences in 2012.
```
dlesnoff, line 9: Who are the Nobel Prize laureates?

stable_marriage/README.md
ZoomRmc: I think they should be kept together, of course, but we'll desync with the [implementations in other languages](https://the-algorithms.com/category/stablemarriage) (there's just two of them there, though, and the Python one lives in the *graphs* category).

If anyone proposes an implementation of one of these algos, we can adapt the directory structure.

dlesnoff (approved):
Ok. We will adapt the directory structure according to the contributions.

PR #62, opened 2023-11-11
feat: add `extended_gcd.nim`

vil02:
This PR adds [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm). The suggested implementation handles negative inputs.

### DIFF
--- a/maths/extended_gcd.nim
+++ b/maths/extended_gcd.nim
@@ -30,6 +30,7 @@ func extendedGCD*(a, b: int): (int, int, int) =
   ## computes their [`gcd`](https://en.wikipedia.org/wiki/Greatest_common_divisor)
   ## and integers `x` and `y`, such that
   ## `gcd = a * x + b * y`
+  ## Note: there are more efficient ways to compute just the `gcd`.
   let (gcd, x, y) = euclidIteration(abs(a), abs(b))
   return (gcd, math.sgn(a) * x, math.sgn(b) * y)
 
--- a/maths/extended_gcd.nim
+++ b/maths/extended_gcd.nim
@@ -1,6 +1,6 @@
 ## Extended Euclidean algorithm
 ##
-## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm)
+## Provides the implementation of the function extendedGCD*
 
 import std/math
 
--- a/maths/extended_gcd.nim
+++ b/maths/extended_gcd.nim
@@ -7,8 +7,8 @@ func updateCoefficients(t0, t1, q: int): (int, int) =
   (t1, t0 - q * t1)
 
 
-func euclidIteration(inA, inB: int): (uint, int, int) =
-  var (a, b) = (inA.abs().uint, inB.abs().uint)
+func euclidIteration(inA, inB: int): (Natural, int, int) =
+  var (a, b) = (inA.abs().Natural, inB.abs().Natural)
   var (x0, x1) = (1, 0)
   var (y0, y1) = (0, 1)
   while b != 0:
@@ -20,7 +20,7 @@ func euclidIteration(inA, inB: int): (uint, int, int) =
   (a, x0, y0)
 
 
-func extendedGCD*(a, b: int): tuple[gcd: uint; x, y: int] =
+func extendedGCD*(a, b: int): tuple[gcd: Natural; x, y: int] =
   ## Implements the
   ## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm).
   ## For given integers `a`, `b` it
@@ -38,20 +38,20 @@ when isMainModule:
   import std/[unittest, sequtils, strformat]
   suite "extendedGCD":
     const testCases = [
-      (10, 15, 5u, -1, 1),
-      (-10, -15, 5u, 1, -1),
-      (32, 64, 32u, 1, 0),
-      (0, 0, 0u, 0, 0),
-      (7, 0, 7u, 1, 0),
-      (-8, 0, 8u, -1, 0),
-      (48, 60, 12u, -1, 1),
-      (98, 56, 14u, -1, 2),
-      (10, -15, 5u, -1, -1),
-      (997, 12345, 1u, -3467, 280),
-      (997, 1234567, 1u, -456926, 369),
+      (10, 15, 5, -1, 1),
+      (-10, -15, 5, 1, -1),
+      (32, 64, 32, 1, 0),
+      (0, 0, 0, 0, 0),
+      (7, 0, 7, 1, 0),
+      (-8, 0, 8, -1, 0),
+      (48, 60, 12, -1, 1),
+      (98, 56, 14, -1, 2),
+      (10, -15, 5, -1, -1),
+      (997, 12345, 1, -3467, 280),
+      (997, 1234567, 1, -456926, 369),
       ].mapIt:
         let tc = (id: fmt"a={it[0]}, b={it[1]}", a: it[0], b: it[1],
-            gcd: it[2], x: it[3], y: it[4])
+            gcd: it[2].Natural, x: it[3], y: it[4])
         if tc.gcd != 0:
           assert tc.a mod int(tc.gcd) == 0
           assert tc.b mod int(tc.gcd) == 0
--- a/maths/extended_gcd.nim
+++ b/maths/extended_gcd.nim
@@ -24,7 +24,7 @@ func absAsUint(x: int): uint =
   uint(abs(x))
 
 
-func extendedGCD*(a, b: int): (uint, int, int) =
+func extendedGCD*(a, b: int): tuple[gcd: uint; x, y: int] =
   ## Implements the
   ## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm).
   ## For given integers `a`, `b` it
--- a/maths/extended_gcd.nim
+++ b/maths/extended_gcd.nim
@@ -35,7 +35,7 @@ func extendedGCD*(a, b: int): (uint, int, int) =
   ## - if `a != 0`, then `abs(y) <= abs(a div gcd)`,
   ## - if `b != 0`, then `abs(x) <= abs(b div gcd)`.
   let (gcd, x, y) = euclidIteration(absAsUint(a), absAsUint(b))
-  return (gcd, math.sgn(a) * x, math.sgn(b) * y)
+  (gcd: gcd, x: math.sgn(a) * x, y: math.sgn(b) * y)
 
 
 when isMainModule:
--- a/maths/extended_gcd.nim
+++ b/maths/extended_gcd.nim
@@ -6,60 +6,61 @@ import std/math
 
 
 func euclidIteration(a, b: int): (int, int, int) =
-    assert a >= 0 and b >= 0
-    if b == 0:
-        return (a, (if a == 0: 0 else: 1), 0)
+  assert a >= 0 and b >= 0
+  if b == 0:
+    (a, (if a == 0: 0 else: 1), 0)
+  else:
     let (gcd, x, y) = euclidIteration(b, a mod b)
-    return (gcd, y, x - (a div b) * y)
+    (gcd, y, x - (a div b) * y)
 
 
 func extendedGCD*(a, b: int): (int, int, int) =
-    ## Implements the
-    ## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm).
-    ## For given integers `a`, `b` it
-    ## computes their [`gcd`](https://en.wikipedia.org/wiki/Greatest_common_divisor)
-    ## and integers `x` and `y`, such that
-    ## `gcd = a*x + b*y`
-    let (gcd, x, y) = euclidIteration(abs(a), abs(b))
-    return (gcd, math.sgn(a)*x, math.sgn(b)*y)
+  ## Implements the
+  ## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm).
+  ## For given integers `a`, `b` it
+  ## computes their [`gcd`](https://en.wikipedia.org/wiki/Greatest_common_divisor)
+  ## and integers `x` and `y`, such that
+  ## `gcd = a * x + b * y`
+  let (gcd, x, y) = euclidIteration(abs(a), abs(b))
+  return (gcd, math.sgn(a) * x, math.sgn(b) * y)
 
 
 when isMainModule:
-    import std/[unittest, sequtils]
-    import std/strformat
-    suite "extendedGCD":
-        const testCases = [
-          (10, 15, 5, -1, 1),
-          (-10, -15, 5, 1, -1),
-          (32, 64, 32, 1, 0),
-          (0, 0, 0, 0, 0),
-          (7, 0, 7, 1, 0),
-          (-8, 0, 8, -1, 0),
-          (48, 60, 12, -1, 1),
-          (98, 56, 14, -1, 2),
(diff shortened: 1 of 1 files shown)


dlesnoff:
I enjoyed this well-tested PR very much. Some points that need improvement:
- [ ] Replace the generic variable names. We have names for them: dividend, divisor, quotient, remainder.
- [ ] Remove the ternary operator. It may look nice, but I expect the first implementation in the file to be understandable by a first-year CS student who only knows a subset of the language.

Please consider proposing an alternative iterative implementation. The proposed implementation is very close to the mathematical process, with triplets as return values. I expect an iterative implementation to provide a more efficient C code. If dealing with signed integers is too complex in the iterative version, restrict yourself to positive values and add checks (or use Positive types).

This algorithm is crucial to compute the inverse of an element over $\mathbb{Z}/n\mathbb{Z}$ (modular ring) or (polynomial ring) $\mathbb{K}[X]$. You may update the description accordingly.

There is no description of the algorithm, just Wikipedia references. I would enjoy a basic description of the input/output.
The complexity of the algorithm is quadratic in the bitsize of the dividend.

Propose links to other algorithms! If we only want to compute the greatest common divisor then the binary Euclidean algorithm gets more practical.

ZoomRmc:
> * Remove the ternary operator. It may look nice, but I expect the first implementation in the file to be understandable by a first-year CS student who only knows a subset of the language.

If we're talking about line 11, I wouldn't be so strict there. First of all, this is not a ternary operator but a full-fledged if-expression, moreover, it's prudently parenthesized. To me, it reads pretty self-explanatory, and I don't have a finished CS degree.

maths/extended_gcd.nim
```diff
@@ -0,0 +1,65 @@
...
+
+import std/math
+
+
+func euclidIteration(a, b: int): (int, int, int) =
+    assert a >= 0 and b >= 0
+    if b == 0:
+        return (a, (if a == 0: 0 else: 1), 0)
+    let (gcd, x, y) = euclidIteration(b, a mod b)
+    return (gcd, y, x - (a div b) * y)
```
ZoomRmc, line 13:
```suggestion
    if b == 0:
      (a, (if a == 0: 0 else: 1), 0)
    else:
      let (gcd, x, y) = euclidIteration(b, a mod b)
      (gcd, y, x - (a div b) * y)
```
You can also leave the first early return and do without the else branch but I'd say the suggested version is preferable as it's immediately obvious that the second block is only valid when the condition isn't triggered. 

See [Result return]( https://github.com/TheAlgorithms/Nim/blob/main/CONTRIBUTING.md#result-return) part of the guidelines.

maths/extended_gcd.nim
```diff
@@ -0,0 +1,65 @@
...
+
+func extendedGCD*(a, b: int): (int, int, int) =
+    ## Implements the
+    ## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm).
+    ## For given integers `a`, `b` it
+    ## computes their [`gcd`](https://en.wikipedia.org/wiki/Greatest_common_divisor)
+    ## and integers `x` and `y`, such that
+    ## `gcd = a*x + b*y`
+    let (gcd, x, y) = euclidIteration(abs(a), abs(b))
+    return (gcd, math.sgn(a)*x, math.sgn(b)*y)
```
ZoomRmc, line 24:
```suggestion
    return (gcd, math.sgn(a) * x, math.sgn(b) * y)
```
Inconsistent spacing around math operators.

maths/extended_gcd.nim
```diff
@@ -0,0 +1,65 @@
...
+          (997, 1234567, 1, -456926, 369),
+            ].mapIt:
+                let tc = (id: fmt"a={it[0]}, b={it[1]}", a: it[0], b: it[1],
+                        gcd: it[2], x: it[3], y: it[4])
+                if tc.gcd != 0:
+                    assert tc.a mod tc.gcd == 0
+                    assert tc.b mod tc.gcd == 0
+                else:
+                    assert tc.a == 0 and tc.b == 0
+                assert tc.gcd == tc.a*tc.x + tc.b*tc.y
```
ZoomRmc, line 51:
```suggestion
                assert tc.gcd == tc.a * tc.x + tc.b * tc.y
```
Inconsistent spacing around math operators.

ZoomRmc (changes requested):
* [ ] Nim idiomatic code uses 2-space identation. See [the official style guide](https://nim-lang.org/docs/nep1.html#style-guidelines-spacing-and-whitespace-conventions), this project follows it in this regard.

Other than the mentioned small issues, looks good. The way you deal with preparing the test cases is an exemplary pattern.

Thanks for the contribution!

vil02:
> > * Remove the ternary operator. It may look nice, but I expect the first implementation in the file to be understandable by a first-year CS student who only knows a subset of the language.
> 
> If we're talking about line 11, I wouldn't be so strict there. First of all, this is not a ternary operator but a full-fledged if-expression, moreover, it's prudently parenthesized. To me, it reads pretty self-explanatory, and I don't have a finished CS degree.

I haven't been studying CS as well.

@ZoomRmc thanks for spotting the formatting issue. It seems that I put too much trust into `nimpretty`. Is there away, to _tell_ `nimpretty` about these rules? (i.e. indent size and space around operators). 

@dlesnoff I will need some time to address all of your remarks.

vil02:
@dlesnoff:
- d534e55820b85e859785268f7a455455aa354521 - the  _iterative approach_ is significantly faster. Running the computation for all inputs between 1 and 1000 is about _twice_ fast (in both cases I used `-d:release`).
- (this will be continued)

dlesnoff:
> the iterative approach is significantly faster. 

Cool!

> Regarding the variable naming, I am not sure if having something like extendedGCD*(dividend, divisor: int) makes things any easier - by nature the gcd is symmetric.

You are right! You wrote in the docstring which temporary variable goes with which variable. That's the only thing that matters.

> the Bézout coefficients can still be negative.

Indeed! No need to write the function for unsigned entries only. I was thinking about inversion for modular arithmetic.

> extended Euclidean algorithm has many applications - that is why I am adding it. I am not sure if it makes sense to mention them

It makes sense to mention them to remember to add a reference in the future to other implementations in the project.
In particular, finding an inverse in a quotient ring is the principal use case of computing Bézout's coefficients that I know of.

> (out of scope for this PR) adding a generic version of the extended Euclidean algorithm working for inputs of some Euclidean domain would be a nice project

For performance reasons, we implement the algorithm for the Euclidean domain that we are working with. We do not aim to implement a linear algebra library here!

maths/extended_gcd.nim
```diff
@@ -0,0 +1,76 @@
...
+
+
+func extendedGCD*(a, b: int): (int, int, int) =
+  ## Implements the
+  ## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm).
+  ## For given integers `a`, `b` it
+  ## computes their [`gcd`](https://en.wikipedia.org/wiki/Greatest_common_divisor)
+  ## and integers `x` and `y`, such that
+  ## `gcd = a * x + b * y`
+  ## Note: there are more efficient ways to compute just the `gcd`.
```
dlesnoff, line 33: I would remove the note. It is too vague (you do not cite these ways) and finally, pretty obvious to the initiated.

maths/extended_gcd.nim
```diff
@@ -0,0 +1,76 @@
...
+  (a, x0, y0)
+
+
+func extendedGCD*(a, b: int): (int, int, int) =
+  ## Implements the
+  ## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm).
+  ## For given integers `a`, `b` it
+  ## computes their [`gcd`](https://en.wikipedia.org/wiki/Greatest_common_divisor)
+  ## and integers `x` and `y`, such that
+  ## `gcd = a * x + b * y`
```
dlesnoff, line 32: In which range lie the Bézout's coefficients that you return?

maths/extended_gcd.nim
```diff
@@ -0,0 +1,76 @@
+## Extended Euclidean algorithm
+##
+## Provides the implementation of the function extendedGCD*
```
dlesnoff, line 3: I would remove this sentence. It's a truism.

maths/extended_gcd.nim
```diff
@@ -0,0 +1,76 @@
...
+
+import std/math
+
+
+func updateCoefficients(t0, t1, q: int): (int, int) =
+  (t1, t0 - q * t1)
+
+
+func euclidIteration(inA, inB: int): (int, int, int) =
+  assert inA >= 0 and inB >= 0
```
dlesnoff, line 13: If they should be positive, why use `int` here?

maths/extended_gcd.nim
```diff
@@ -0,0 +1,76 @@
+## Extended Euclidean algorithm
+##
+## Provides the implementation of the function extendedGCD*
+
+import std/math
+
+
+func updateCoefficients(t0, t1, q: int): (int, int) =
+  (t1, t0 - q * t1)
```
dlesnoff, line 9: Is this function really necessary? I like putting in a function something that appears twice at the only condition that the name of the function is explicit. This is not the case here. What kind of update does it perform?

dlesnoff (changes requested):
Some minor changes (documentation) are required. Very good tests.

maths/extended_gcd.nim
vil02: Done in 5e538a26bf1d9615fcc7df464c57d46f59dd6216.

maths/extended_gcd.nim
vil02: Please have a look at 1faa670bef060aa73fa61ef43ebd02949cf79d70. The advantage is, that the resulting `gcd` can then also be a `uint`. The disadvantage:
- all of the _ugly casts_ - not so ugly as I thought at first,
- possible overflow in the future - ok, now `euclidIteration` is just called from `extendedGCD*`.

Please let me know about your opinion here.

maths/extended_gcd.nim
vil02: Yes, it is - it shows (as clearly as possible) that the `x`'s and `y`'s evolve in the same way.

I am also not entirely happy with the name. Do you have some suggestions?

maths/extended_gcd.nim
vil02: cf. fdfebe6d6a726148a3985c6606735cd9fb304c53

maths/extended_gcd.nim
vil02: cf. 6d93679ddbba5c778844e0299b4a198fd98e42e0

vil02:
> > extended Euclidean algorithm has many applications - that is why I am adding it. I am not sure if it makes sense to mention them
> 
> It makes sense to mention them to remember to add a reference in the future to other implementations in the project. In particular, finding an inverse in a quotient ring is the principal use case of computing Bézout's coefficients that I know of.
> 

I think everyone can find easily where this function will be used. If they are interested about applications of extended Euclidean algorithm, they would look into some textbook or... wikipedia.

> For performance reasons, we implement the algorithm for the Euclidean domain that we are working with. We do not aim to implement a linear algebra library here!

(off-topic) I would not use the word _linear_ here.

maths/extended_gcd.nim
dlesnoff: Two casts at each call are better than this assert. This raises an AssertionDefect at runtime and the user doesn't know why. With types, the error is caught at compile-time.

maths/extended_gcd.nim
```diff
@@ -0,0 +1,90 @@
...
+  ## and furthermore:
+  ## - if `a != 0`, then `abs(y) <= abs(a div gcd)`,
+  ## - if `b != 0`, then `abs(x) <= abs(b div gcd)`.
+  let (gcd, x, y) = euclidIteration(absAsUint(a), absAsUint(b))
+  return (gcd, math.sgn(a) * x, math.sgn(b) * y)
+
+
+when isMainModule:
+  import std/[unittest, sequtils]
+  import std/strformat
```
ZoomRmc, line 43:
```suggestion
  import std/[unittest, sequtils, strformat]
```

ZoomRmc (changes requested):
Could you please explain your choice of `uint` for the `gcd` value? Does the algorithm require it? As far as I can say after a quick review, the public proc takes a couple of `int` values, surely their GCD can't be of a bigger magnitude then either of them?

Nicely done, sorry for the belated review.

maths/extended_gcd.nim
```diff
@@ -0,0 +1,90 @@
...
+  ## [Extended Euclidean algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm).
+  ## For given integers `a`, `b` it
+  ## computes their [`gcd`](https://en.wikipedia.org/wiki/Greatest_common_divisor)
+  ## and integers `x` and `y`, such that
+  ## `gcd = a * x + b * y`,
+  ## and furthermore:
+  ## - if `a != 0`, then `abs(y) <= abs(a div gcd)`,
+  ## - if `b != 0`, then `abs(x) <= abs(b div gcd)`.
+  let (gcd, x, y) = euclidIteration(absAsUint(a), absAsUint(b))
+  return (gcd, math.sgn(a) * x, math.sgn(b) * y)
```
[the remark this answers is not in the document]
vil02: I prefer `math.sgn`. Especially, when the code is already written.

maths/extended_gcd.nim
```diff
@@ -0,0 +1,90 @@
...
+    let q = int(a div b)
+    (x0, x1) = updateCoefficients(x0, x1, q)
+    (y0, y1) = updateCoefficients(y0, y1, q)
+    (a, b) = (b, a mod b)
+
+  (a, x0, y0)
+
+
+func absAsUint(x: int): uint =
+  uint(abs(x))
```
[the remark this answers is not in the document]
vil02: Done in 20f7366e781a4e768ea946f6f52b56120b15f421.

vil02:
> Could you please explain your choice of `uint` for the `gcd` value? Does the algorithm require it? As far as I can say after a quick review, the public proc takes a couple of `int` values, surely their GCD can't be of a bigger magnitude then either of them?

At the _very beginning_ (cf. 34a4ae234926461d2d7303cc1f377c3ce1c2396b) it was an `int`.

When `gcd` is of the type `uint` then it is clear that `gcd` is never negative (I always find it very confusing what could be `gcd(-4, -2)` or `gcd(4, -2)`). Clearly I would prefer a type `nonngetiveIntWhichCanBeCastedToIntWithoutOverflow`, but I think Nim does not provide such type.

ZoomRmc:
> Clearly I would prefer a type `nonngetiveIntWhichCanBeCastedToIntWithoutOverflow`, but I think Nim does not provide such type.

Nim has [range types](https://nim-lang.org/docs/manual.html#types-subrange-types), and in particular [`Positive`](https://nim-lang.org/docs/system.html#Positive), which can be used anywhere where `int` is taken.

While the topic is highly debatable, using unsigned integers in cases where overflow behaviour or bit-manipulation isn't required is usually frowned upon and considered out-of-style. There were some rather lengthy discussion on the forums and in GH issues, but I can't provide links right now.

dlesnoff:
There are [many problems with range types](https://status-im.github.io/nim-style-guide/language.range.html) but Positive is clearly what @vil02 is looking for and it should be ok for implementations in the Algorithms repository.
I thought I mentioned this earlier in the discussion, sorry.

vil02:
`0` is not `Positive`. Note that `gcd(0, 0)` is `0`.

ZoomRmc:
> `0` is not `Positive`. Note that `gcd(0, 0)` is `0`.

Oh, right, `Natural` then.

ZoomRmc (approved):

dlesnoff (approved):
Well done!

PR #63, opened 2023-11-14
feat: add Fibonacci Numbers

Archargelod:
Low-hanging fruit. And it's also on the front page of the website. 😉

### DIFF
--- a/dynamic_programming/fibonacci_numbers.nim
+++ b/dynamic_programming/fibonacci_numbers.nim
@@ -19,7 +19,7 @@
 
 runnableExamples:
 
-  doAssert fibonacciSeqIterative(5.Natural) == @[Natural(0), 1, 1, 2, 3]
+  doAssert fibonacciSeqIterative(5.Positive) == @[Natural(0), 1, 1, 2, 3]
 
   doAssert fibonacciClosure(4.Natural) == Natural(3)
 
@@ -50,7 +50,7 @@ proc fibonacciClosure*(nth: Natural): Natural =
     discard fib()
   fib()
 
-proc fibonacciSeqClosure*(n: Natural): seq[Natural] =
+proc fibonacciSeqClosure*(n: Positive): seq[Natural] =
   ## Generates a list of n first fibonacci numbers with use of a closure.
   result = newSeq[Natural](n)
   result[0] = 0
@@ -60,7 +60,7 @@ proc fibonacciSeqClosure*(n: Natural): seq[Natural] =
     for i in 2..<n:
       result[i] = fib()
 
-func fibonacciSeqIterative*(n: Natural): seq[Natural] =
+func fibonacciSeqIterative*(n: Positive): seq[Natural] =
   ## Generates a list of n first fibonacci numbers in iterative manner.
   result = newSeq[Natural](n)
   result[0] = 0
--- a/dynamic_programming/fibonacci_numbers.nim
+++ b/dynamic_programming/fibonacci_numbers.nim
@@ -48,7 +48,8 @@ func pow(matrix: fibMatrix, n: Natural): fibMatrix =
 func fibonacciMatrix*(nth: Natural): Natural =
   ## Calculates the n-th fibonacci number with use of matrix arithmetic.
   if nth <= 1: return nth
-  var matrix = [[1, 1], [1, 0]]
+  var matrix = [[1, 1],
+                [1, 0]]
   matrix.pow(nth - 1)[0][0]
 
 func fibonacciRecursive*(nth: Natural): Natural =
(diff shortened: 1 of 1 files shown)


dynamic_programming/fibonacci_numbers.nim
```diff
@@ -0,0 +1,152 @@
...
+  for i in 0..s.b:
+    if i <= 1:
+      if i >= s.a: yield i
+    else:
+      swap(prev, current)
+      current += prev
+      if i >= s.a: yield current
+
+func fibonacciBinet*(nth: Natural): Natural =
+  ## Calculates the n-th fibonacci number in constant time O(1) with use of Binet's formula.
```
dlesnoff, line 83: *Approximates. Binet's formula has only been rediscovered by Binet (It was previously known by Euler and even maybe by De Moivre). I would remove this nomination and add a reference to the formula in the docstring.

dlesnoff:
Thanks for this new contribution that I am going to read with pleasure in detail this week-end. This contribution seems more suited to showcase the language features than the algorithms themselves.

Have you considered adding the matrix construction to these Fibonacci procedures?
The idea is that if you compute the matrix given by writing out the vector F_n+2 and F_n+1 in terms of a matrix operation with the vector F_n+1 and F_n, you can compute e.g. the n-th term in O(MM(n) log(n)) operations using binary exponentiation (MM(n) is the cost of multiplication of matrices of size n).
Here, there are very few values that can be stored on an unsigned integer for the Fibonacci sequence but this idea is general and can be applied to any linear recurrence sequence.

I might write this in a Fiduccia-specific algorithm implementation but I wonder if we could not simply give an example here of the advantages of the matrix representation.

dynamic_programming/fibonacci_numbers.nim
Archargelod: - [x] Ok. Besides, it's not really an algorithm in the traditional sense. I'll remove it.

Archargelod:
> Have you considered adding the matrix construction to these Fibonacci procedures?
I think I've seen the implementation of it in other languages, but have never attempted to write one myself. 
I will try to explore this idea and see if I can come up with something before the weekend.

dynamic_programming/fibonacci_numbers.nim
dlesnoff: I would keep the algorithm, it it is educational to compare the Fibonacci numbers with the approximation. I would just change the name. The description is misleading because we can not compute Fibonacci numbers exactly in O(1). The formula requires infinite precision to output the exact value.

Archargelod:
Not sure if it's the optimal matrix version for Fibonacci numbers or the correct one. Perhaps it could more general in some places. But there's only so much a person could learn in a couple hours on the internet, filling a void left out after the "educational facilities".

And it also works. 😄

dynamic_programming/fibonacci_numbers.nim
Archargelod: > Binet's formula has only been rediscovered by Binet...

I used name of Binet here, because wikipedia says:
> the Fibonacci numbers have a closed-form expression. It has become known as Binet's formula.

What names should I use instead?
Currently, the procedure is `fibonacciBinet` and in the tests I refer to it as "Constant time formula".
Are they both should be changed?

dlesnoff: The French wikipedia page explains that it has been rediscovered by Binet.

I would give them a descriptive name: `fibonacciClosedFormApproximation`

dynamic_programming/fibonacci_numbers.nim
```diff
@@ -0,0 +1,186 @@
...
+    b = m1[0][0] * m2[0][1] + m1[0][1] * m2[1][1]
+    z = m1[1][0] * m2[0][0] + m1[1][1] * m2[1][0]
+    y = m1[1][0] * m2[0][1] + m1[1][1] * m2[1][1]
+
+  [[a, b], [z, y]]
+
+func pow(matrix: fibMatrix, n: Natural): fibMatrix =
+  ## binary matrix exponentiation (divide-and-conquer approach)
+  if n in {0, 1}:
+    matrix
```
dlesnoff, line 41: This looks suspicious to me. If $n == 0$, you want to return the identity matrix instead.

dynamic_programming/fibonacci_numbers.nim
```diff
@@ -0,0 +1,186 @@
...
+
+  doAssert fibonacciClosure(4) == Natural(3)
+
+  for f in fibonacciIterator(1.Natural..4.Natural):
+    echo f # 1, 1, 2, 3
+
+import std/math
+
+type
+  fibMatrix = array[2, array[2, int]]
```
dlesnoff, line 27: Matrix2x2 would be a better name. The multiplication and exponentiation algorithms below would work for more general matrices.

dynamic_programming/fibonacci_numbers.nim
```diff
@@ -0,0 +1,186 @@
...
+
+func `*`(m1, m2: fibMatrix): fibMatrix =
+  let
+    a = m1[0][0] * m2[0][0] + m1[0][1] * m2[1][0]
+    b = m1[0][0] * m2[0][1] + m1[0][1] * m2[1][1]
+    z = m1[1][0] * m2[0][0] + m1[1][1] * m2[1][0]
+    y = m1[1][0] * m2[0][1] + m1[1][1] * m2[1][1]
```
dlesnoff, line 34: Why did you unroll the for loop here?

dynamic_programming/fibonacci_numbers.nim
```diff
@@ -0,0 +1,186 @@
...
+    halfPow * halfPow
+  else:
+    matrix * matrix.pow(n-1)
+
+func fibonacciMatrix*(nth: Natural): Natural =
+  ## Calculates the n-th fibonacci number with use of matrix arithmetic.
+  if nth <= 1: return nth
+  var matrix = [[1, 1],
+                [1, 0]]
+  matrix.pow(nth - 1)[0][0]
```
dlesnoff, line 53: Cool! The initial matrix could be a const and put at the top of the file.

dynamic_programming/fibonacci_numbers.nim
```diff
@@ -0,0 +1,186 @@
...
+  if n in {0, 1}:
+    matrix
+  elif n mod 2 == 0:
+    let halfPow = matrix.pow(n div 2)
+    halfPow * halfPow
+  else:
+    matrix * matrix.pow(n-1)
+
+func fibonacciMatrix*(nth: Natural): Natural =
+  ## Calculates the n-th fibonacci number with use of matrix arithmetic.
```
dlesnoff, line 49: There is also an algorithm not much more complex using only «polynomial» arithmetic that could be a good follow-up.

dlesnoff (changes requested):
I'll give it a closer look next week-end.

I would separate with comments the different sorts of implementations with the easier-to-grasp functions at the top.

E.g.:
```nim
## Recursive implementations
## List of first terms
## Closure and iterators
## An asymptotic faster matrix algorithm
(## An asymptotic faster polynomial algorithm) <- I'll add this one after!
```

dynamic_programming/fibonacci_numbers.nim
Archargelod: Ok.

dynamic_programming/fibonacci_numbers.nim
Archargelod: Done.

dynamic_programming/fibonacci_numbers.nim
Archargelod: Done.

dynamic_programming/fibonacci_numbers.nim
Archargelod: it looks a lot more clearer to me than:
```Nim
  for i in 0..1:
    for j in 0..1:
      for k in 0..1:
        result[i][j] += m1[i][k] * m2[k][j]
```
And is quite a bit faster, at least in micro benchmarks.

dynamic_programming/fibonacci_numbers.nim
dlesnoff: How come this change had no impact on the tests?
I am surprised that tests passed before.

Archargelod: I guess, because the only case where n == 0 in this algorithm is F(1) and the Initial Fibonacci matrix [0][0] coincidentally matches Identity Matrix [0][0].

Oh, and I also just skip 0 and 1, because they're trivial in fibonacci sequence, so F(1) never even calls the matrix implementation.
```Nim
  if nth <= 1: return nth
```

dlesnoff: This doesn't explain why the result of pow was correct.
All calls to pow eventually call pow(0) or pow(1).

Why are you performing binary exponentiation recursively?
You are adding all intermediate matrices to the call stack.
You can look at: https://github.com/jilljenn/tryalgo/blob/master/tryalgo/fast_exponentiation.py
for two iterative versions of the algorithm.
Remove the asserts, the modulo ops., make it Nimesque and apply it to your matrix case.
Notice that if $n$ is odd, then $n-1$ is necessary even, so you can combine in this case the two operations to reduce the number of recursive calls.

Archargelod: Thank you for the feedback and the link.
Honestly, I had to grab a pen and paper to walk through the algorithm step by step, but now, I think, your last sentence is much clearer to me.

I've added the second algorithm for iterative pow from your link and also a couple tests for the exponentiation function.

dlesnoff (approved):
Nice job!
Many thanks for adding the matrix algorithm.
I tried implementing the polynomial version, but I prefer making something more general for linear recurrence sequences, as it is not faster than the matrix algorithm for order-2 integer sequences like Fibonacci.

PR #66, opened 2023-12-18
feat: add `modular_inverse.nim`

vil02:
After #62 being merged, the most natural step is to add a function computing the modular inverse.

### DIFF
--- a/maths/modular_inverse.nim
+++ b/maths/modular_inverse.nim
@@ -3,7 +3,7 @@
 import std/[options, math]
 
 
-func euclidHalfIteration(inA, inB: Positive): (Natural, int) =
+func euclidHalfIteration(inA, inB: Positive): tuple[gcd: Natural, coeff: int] =
   var (a, b) = (inA.Natural, inB.Natural)
   var (x0, x1) = (1, 0)
   while b != 0:
--- a/maths/modular_inverse.nim
+++ b/maths/modular_inverse.nim
@@ -32,7 +32,9 @@ when isMainModule:
   suite "modularInverse":
     const testCases = [
       (3, 7, 5),
-      (-1, 5, 4),
+      (-1, 5, 4), # Inverse of a negative
+      (-7, 5, 2), # Inverse of a negative lower than modulus
+      (-7, 4, 1), # Inverse of a negative with non-prime modulus
       (4, 5, 4),
       (9, 5, 4),
       (5, 21, 17),
--- a/maths/modular_inverse.nim
+++ b/maths/modular_inverse.nim
@@ -1,0 +1,56 @@
+## Modular inverse
+
+import std/[options, math]
+
+import extended_gcd
+
+
+func modularInverse*(a: int, modulus: Natural): Option[Positive] =
+  ## For given integer `a` and a natural number `modulus` it
+  ## computes the inverse of `a` modulo `modulus`, i.e.
+  ## it finds an integer `0 < inv < modulus` such that
+  ## `(a * inv) mod modulus == 1`.
+  if modulus == 1:
+    return none(Positive)
+  let (gcd, x, _) = extended_gcd.extendedGCD(a, modulus)
+  if gcd == 1:
+    return some(math.floorMod(x, modulus).Positive)
+  none(Positive)
+
+
+when isMainModule:
+  import std/[unittest, sequtils, strformat]
+  suite "modularInverse":
+    const testCases = [
+      (3, 7, 5),
+      (-1, 5, 4),
+      (4, 5, 4),
+      (9, 5, 4),
+      (5, 21, 17),
+      (2, 21, 11),
+      (4, 21, 16),
+      (55, 372, 115),
+      (1, 100, 1),
+      ].mapIt:
+        let tc = (id: fmt"a={it[0]}, modulus={it[1]}", a: it[0], modulus: it[1],
+            inv: it[2])
+        assert 0 < tc.inv
+        assert tc.inv < tc.modulus
+        assert math.floorMod(tc.a * tc.inv, tc.modulus) == 1
+        tc
+
+    for tc in testCases:
+      test tc.id:
+        checkpoint("returns expected result")
+        check modularInverse(tc.a, tc.modulus).get() == tc.inv
+
+    test "No inverse when modulus is 1":
+      check modularInverse(0, 1).is_none()
+      check modularInverse(1, 1).is_none()
+      check modularInverse(-1, 1).is_none()
+
+    test "No inverse when inputs are not co-prime":
+      check modularInverse(2, 4).is_none()
+      check modularInverse(-5, 25).is_none()
+      check modularInverse(0, 17).is_none()
+      check modularInverse(17, 17).is_none()
--- a/maths/modular_inverse.nim
+++ b/maths/modular_inverse.nim
@@ -8,7 +8,7 @@ func euclidHalfIteration(inA, inB: Positive): (Natural, int) =
   var (x0, x1) = (1, 0)
   while b != 0:
     (x0, x1) = (x1, x0 - (a div b) * x1)
-    (a, b) = (b, a mod b)
+    (a, b) = (b, math.floorMod(a, b))
 
   (a, x0)
 
--- a/maths/modular_inverse.nim
+++ b/maths/modular_inverse.nim
@@ -2,8 +2,6 @@
 
 import std/[options, math]
 
-import ./extended_gcd.nim
-
 
 func euclidHalfIteration(inA, inB: Positive): (Natural, int) =
   var (a, b) = (inA.Natural, inB.Natural)
--- a/maths/modular_inverse.nim
+++ b/maths/modular_inverse.nim
@@ -2,17 +2,27 @@
 
 import std/[options, math]
 
-import extended_gcd
+import ./extended_gcd.nim
 
 
-func modularInverse*(a: int, modulus: Natural): Option[Positive] =
+func euclidHalfIteration(inA, inB: Natural): (Natural, int) =
+  var (a, b) = (inA, inB)
+  var (x0, x1) = (1, 0)
+  while b != 0:
+    (x0, x1) = (x1, x0 - (a div b) * x1)
+    (a, b) = (b, a mod b)
+
+  (a, x0)
+
+
+func modularInverse*(a: int, modulus: Positive): Option[Positive] =
   ## For given integer `a` and a natural number `modulus` it
   ## computes the inverse of `a` modulo `modulus`, i.e.
   ## it finds an integer `0 < inv < modulus` such that
   ## `(a * inv) mod modulus == 1`.
   if modulus == 1:
     return none(Positive)
-  let (gcd, x, _) = extended_gcd.extendedGCD(a, modulus)
+  let (gcd, x) = euclidHalfIteration(math.floorMod(a, modulus), modulus)
   if gcd == 1:
     return some(math.floorMod(x, modulus).Positive)
   none(Positive)
--- a/maths/modular_inverse.nim
+++ b/maths/modular_inverse.nim
@@ -28,7 +28,7 @@ func modularInverse*(inA: int, modulus: Positive): Option[Positive] =
 
 
 when isMainModule:
-  import std/[unittest, sequtils, strformat]
+  import std/[unittest, sequtils, strformat, random]
   suite "modularInverse":
     const testCases = [
       (3, 7, 5),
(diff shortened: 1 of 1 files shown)


maths/modular_inverse.nim
```diff
@@ -0,0 +1,56 @@
+## Modular inverse
+
+import std/[options, math]
+
+import extended_gcd
```
vil02, line 5: Should I change it to
```suggestion
import "extended_gcd.nim"
```
What is the desired way to import _own_ modules?

maths/modular_inverse.nim
```diff
@@ -0,0 +1,56 @@
...
+      (4, 5, 4),
+      (9, 5, 4),
+      (5, 21, 17),
+      (2, 21, 11),
+      (4, 21, 16),
+      (55, 372, 115),
+      (1, 100, 1),
+      ].mapIt:
+        let tc = (id: fmt"a={it[0]}, modulus={it[1]}", a: it[0], modulus: it[1],
+            inv: it[2])
```
vil02, line 36: This indent was introduced by `nim pretty`, when breaking the long line above. It also accepts the indent of _size_ 2.

maths/modular_inverse.nim
dlesnoff: https://nim-lang.org/docs/manual.html#modules-module-names-in-imports
In short:
- std for stdllib packages.
- pkg for external packages (dependencies only).
- ../module_name for modules in own repository.

dlesnoff:
This is a too naïve way of computing the modular inverse of an element $a$ in $\mathbb{Z}/n\mathbb{Z}$.
I know two methods for this:
- The Extended Euclidean algorithm, but please, without computing the extra Bézout's coefficient (in $\mathbb{Z}/n\mathbb{Z}$).
- The [Euler's theorem](https://en.wikipedia.org/wiki/Euler%27s_theorem) or (rather [Fermat's little theorem](https://en.wikipedia.org/wiki/Fermat%27s_little_theorem) in this case) with fast exponentiation for inversion modulo a prime number p, e.g. in Z/pZ.
As a good reference for this:
> Modern Computer Algebra, (Chapter 4 in the 1st edition)

maths/modular_inverse.nim
```diff
@@ -0,0 +1,65 @@
...
+
+import std/[options, math]
+
+
+func euclidHalfIteration(inA, inB: Positive): (Natural, int) =
+  var (a, b) = (inA.Natural, inB.Natural)
+  var (x0, x1) = (1, 0)
+  while b != 0:
+    (x0, x1) = (x1, x0 - (a div b) * x1)
+    (a, b) = (b, a mod b)
```
dlesnoff, line 11:
```suggestion
    (a, b) = (b, math.floorMod(a, b))
```
I think that math.floorMod might be more suited here than Nim's mod operator which may return a negative integer.
You can then return two Naturals.
Another maybe less expensive trick to ensure that `x0` is positive, is to perform an addition in the end:
(x0 + a mod a)

maths/modular_inverse.nim
```diff
@@ -0,0 +1,65 @@
+## Modular inverse
+
+import std/[options, math]
+
+
+func euclidHalfIteration(inA, inB: Positive): (Natural, int) =
```
dlesnoff, line 6:
```suggestion
func euclidHalfIteration(inA, inB: Positive): (gcd: Natural, coeff: Natural) =
```
I prefer naming the return values. This is visible in the documentation and thus in the API exposed to the user.
The names of the resulting variables are not displayed otherwise.

maths/modular_inverse.nim
```diff
@@ -0,0 +1,65 @@
...
+  while b != 0:
+    (x0, x1) = (x1, x0 - (a div b) * x1)
+    (a, b) = (b, a mod b)
+
+  (a, x0)
+
+
+func modularInverse*(inA: int, modulus: Positive): Option[Positive] =
+  ## For given integer `a` and a natural number `modulus` it
+  ## computes the inverse of `a` modulo `modulus`, i.e.
```
dlesnoff, line 18:
```suggestion
  ## For a given integer `a` and a natural number `modulus` it
  ## computes the inverse of `a` modulo `modulus`, i.e.
```

maths/modular_inverse.nim
```diff
@@ -0,0 +1,65 @@
...
+    return some(math.floorMod(x, modulus).Positive)
+  none(Positive)
+
+
+when isMainModule:
+  import std/[unittest, sequtils, strformat]
+  suite "modularInverse":
+    const testCases = [
+      (3, 7, 5),
+      (-1, 5, 4),
```
dlesnoff, line 35:
```suggestion
      (-1, 5, 4), # Inverse of a negative
      (-7, 5, 2), # Inverse of a negative lower than modulus
      (-7, 4, 1), # Inverse of a negative with non-prime modulus
```
These two test cases felt lacking to me. Why not a randomized test here too?

dlesnoff (changes requested):

maths/modular_inverse.nim
vil02: I am not sure what do you mean by `(x0 + a mod a)`. In this situation, `a mod a` is always `0`.

maths/modular_inverse.nim
vil02: Partially done in b7e3dcd85f02c51339f22da1a10e2ad538e656da. As far as I can tell, the suggestion does not have a correct syntax.

maths/modular_inverse.nim
vil02: > These two cases have to be tested as well. Why not a randomized test here too?

I agree that writing some _randomized tests_ here is quite straightforward. I am not sure if that would not be an _overkill_ in this situation. Do you really think that feeding _random_ inputs into this function will help us to discover some bug?

maths/modular_inverse.nim
dlesnoff: I meant $(x_0 + a) \bmod{a}$.

Ensuring that `x0` is positive enables using unsigned arithmetic for large modulus.
The `mod` operator's behaviour is different from the `%` C operator. It returns a result with the sign of its operands which are non-negative in this case.
I would rather regroup everything under the same function `modInvByEuclidsXGCD` and put a comment to distinguish the part that is similar to Euclid's algorithm.
Half-iteration is not a commonly accepted denomination, nor a different algorithm.

maths/modular_inverse.nim
dlesnoff: Thanks. I constantly switch between languages, and I have trouble memorizing the tuple syntax for procedures which I seldom use.

maths/modular_inverse.nim
dlesnoff: Just writing a property-based test to check that the obtained inverse makes 1 when multiplied by the initial element is not an overkill.
It is *pedagogical*.

dlesnoff:
I incited you to consider also reduction using fast exponentiation and Fermat's little theorem or Euler's theorem (in modular arithmetic). I wrote this:
```nim
import std/math

proc modularExponentiation(base: int, exponent: Natural, modulus: Positive): Natural =
  result = 1
  var
    baseMod = math.floorMod(base, modulus)
    exp = exponent

  while exp > 0:
    if math.floorMod(exp, 2) == 1:
      result = math.floorMod((result * baseMod), modulus)
    baseMod = math.floorMod((baseMod * baseMod), modulus)
    exp = exp div 2

proc modularInverseByFermatsLittleTheorem*(a: int, p: Positive): Natural =
  ## Only if p prime
  modularExponentiation(a, p - 2, p)
  # return (result + p) mod p

proc modularInverseByEulersTheorem*(a: int, totient: Positive, p: Positive): Natural =
  ## Computing the totient is as hard as factoring p,
  ## it is thus safer to assume we know the totient.
  modularExponentiation(a, totient-1, p)

when isMainModule:
  import std/[unittest, sequtils, strformat]
  
  # I took your testing suite, but I had to separate cases for
  # prime and composite modulus.
  suite "modularInverse":
    const primeCases = [
      (3, 7, 5),
      (-1, 5, 4), # Inverse of a negative
      (-7, 5, 2), # Inverse of a negative lower than modulus
      (4, 5, 4),
      (9, 5, 4),
      ].mapIt:
        let tc = (id: fmt"a={it[0]}, modulus={it[1]}", a: it[0], modulus: it[1],
            inv: it[2])
        assert 0 < tc.inv
        assert tc.inv < tc.modulus
        assert math.floorMod(tc.a * tc.inv, tc.modulus) == 1
        tc

    const compositeCases = [
      (-7, 2, 4, 1), # Inverse of a negative with non-prime modulus
      (5, 12, 21, 17),
      (2, 12, 21, 11),
      (4, 12, 21, 16),
      (55, 120, 372, 115),
      (1, 40, 100, 1),
      ].mapIt:
        let tc = (id: fmt"a={it[0]}, totient={it[1]}, modulus={it[2]}", a: it[0], totient: it[1], modulus: it[2],
            inv: it[3])
        assert 0 < tc.inv and tc.inv < tc.modulus
        assert 0 < tc.totient and tc.totient < tc.modulus
        assert math.floorMod(tc.a * tc.inv, tc.modulus) == 1
        tc

    for tc in primeCases:
      test tc.id:
        checkpoint("returns expected result")
        check modularInverseByFermatsLittleTheorem(tc.a, tc.modulus) == tc.inv

    for tc in compositeCases:
      test tc.id:
        checkpoint("returns expected result")
        check modularInverseByEulersTheorem(tc.a, tc.totient, tc.modulus) == tc.inv
```

maths/modular_inverse.nim
vil02: > Ensuring that `x0` is positive enables using unsigned arithmetic for large modulus.

During the computation /_iteration_ `x0` still might become negative. Or am I missing something?

> The `mod` operator's behavior is different from the `%` C operator. It returns a result with the sign of its operands which are non-negative in this case.

There is a [good wiki page](https://en.wikipedia.org/wiki/Modulo#In_programming_languages) about that. Basically I _do not trust_ `mod` for negative inputs.

> I would rather regroup everything under the same function `modInvByEuclidsXGCD` and put a comment to distinguish the part that is similar to Euclid's algorithm. Half-iteration is not a commonly accepted denomination, nor a different algorithm.

I prefer to have a separate functions: `euclidHalfIteration` does the computation, `modularInverse*` handles all of the sign madness. This provides more structure than the comments. Similarly as discussed in #62, `euclidHalfIteration` is not a perfect name.

maths/modular_inverse.nim
vil02: I am a bit lost. We have a check if the test case is valid:
https://github.com/TheAlgorithms/Nim/blob/507792656a53ef180f121b490a0ff333ce94ff05/maths/modular_inverse.nim#L50

vil02:
Feel free to _push_ this into my branch (or even better: in a separate PR when this one is merged). Please make sure that `modularInverseByFermatsLittleTheorem*` and [`modularInverse*`](https://github.com/TheAlgorithms/Nim/blob/507792656a53ef180f121b490a0ff333ce94ff05/maths/modular_inverse.nim#L16) (this needs a new name) are tested _together_.

maths/modular_inverse.nim
dlesnoff: I wanted to test the definition of the inverse on randomized inputs. Pseudocode:
```nim
a = randint(0, modulus-1)
inv = modularInverse(a, modulus)
assert a * inv == 1
```

maths/modular_inverse.nim
dlesnoff: I changed my mind. Keep it in two separate functions, `euclidHalfIteration` should be fine.

> During the computation /iteration x0 still might become negative
Could you let me know if you are sure about this? Can you give me two Positive inputs, for which the result of mod is negative?
Rather than the wiki, I checked Nim's official mod documentation in the system module. It says that the result of two positive inputs (quotient and dividend are positive) should return a positive remainder.

dlesnoff:
> Feel free to push this into my branch (or even better: in a separate PR when this one is merged).

I will do it in a separate PR.

maths/modular_inverse.nim
vil02: If you compute `modularInverse(3, 97)` at some point you will reach the state with `x0 = -32`.

maths/modular_inverse.nim
vil02: Please have a look at 2e25c2ad1f0dd14daf14dd58de2b300fa9e98253. The added test does not make any assumptions of the input (e.g. `a` can be bigger than `modulus` or negative). In my opinion, it is better, since we test the _entire implementation_.

[off-topic]: I would highly suggest to change the setup of this project in such a way, that a call `nim test randomTestSize=10000` would be possible. The command `nim test` would run only the _deterministic tests_ and would be used for things like generating coverage reports etc.

dlesnoff (approved):
Thank you for this nice contribution @vil02 !

PR #71, opened 2024-03-15
feat: add `bresenhams_line`

vil02:
This PR adds an implementation of the [Bresenham's line algorithm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm). As always, I am open for all remarks.

### DIFF
--- a/graphics/bresenhams_line.nim
+++ b/graphics/bresenhams_line.nim
@@ -1,0 +1,84 @@
+## Bresenham's line algorithm
+
+func computeStep(inStart, inEnd: int): int =
+  if inStart < inEnd: 1 else: -1
+
+func drawBresenhamLine*(posA, posB: (int, int)): seq[(int, int)] =
+  ## returns a sequence of coordinates approximating the straights line
+  ## between points `posA` and `posB`.
+  ## These points are determined using the
+  ## [Bresenham's line algorithm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm).
+  ## This implementation balances the positive and negative errors between `x` and `y` coordinates.
+  let
+    dx = abs(posA[0]-posB[0])
+    dy = -abs(posA[1]-posB[1])
+    xStep = computeStep(posA[0], posB[0])
+    yStep = computeStep(posA[1], posB[1])
+
+  var
+    difference = dx + dy
+    res: seq[(int, int)] = @[]
+    x = posA[0]
+    y = posA[1]
+
+  res.add((x, y))
+  while (x, y) != posB:
+    let doubleDifference = 2 * difference
+    if doubleDifference >= dy:
+      difference += dy
+      x += xStep
+    if doubleDifference <= dx:
+      difference += dx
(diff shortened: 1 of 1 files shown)


graphics/bresenhams_line.nim
```diff
@@ -0,0 +1,84 @@
...
+      ("line1",
+        (2, 3), (8, 7),
+        @[(2, 3), (3, 4), (4, 4), (5, 5), (6, 6), (7, 6), (8, 7)]),
+      ("line2",
+        (2, 1), (8, 5),
+        @[(2, 1), (3, 2), (4, 2), (5, 3), (6, 4), (7, 4), (8, 5)]),
+      ].mapIt:
+        (name: it[0], posA: it[1], posB: it[2], line: it[3])
+
+    func flipCoordinatesTuple(inPos: (int, int)): (int, int) =
```
ZoomRmc, line 61: I think the idiomatic way is to use overloading and omit the type suffix in the function's name.

graphics/bresenhams_line.nim
```diff
@@ -0,0 +1,84 @@
...
+      (inPos[1], inPos[0])
+
+    func flipCoordinates(line: seq[(int, int)]): seq[(int, int)] =
```
ZoomRmc, line 64:
```suggestion
    func flipCoordinates(line: openArray[(int, int)]): seq[(int, int)] =
```

graphics/bresenhams_line.nim
```diff
@@ -0,0 +1,84 @@
...
+      line.map(flipCoordinatesTuple)
+
+    func reverseLine(line: seq[(int, int)]): seq[(int, int)] =
+      toSeq(reversed(line))
```
ZoomRmc, line 68:
```suggestion
    func reverseLine(line: openArray[(int, int)]): seq[(int, int)] =
      reversed(line)
```
`reversed` already returns a new `seq`.

ZoomRmc: Not sure if this alias is really necessary, just calling `reversed` should be sufficient, especially if the `line`/`path` type were to be defined and used instead of a sequence.

ZoomRmc (changes requested):
A couple of nitpicks:

* Consider renaming `flip` to `swap`, as the former can pose a question if symmetry is involved somehow.
* Consider renaming `line` to something like `path` in the testing routines, as lines are usually defined as a mathematical concept with just two points, but here we have sequences of specific discrete points.
* The code is trivial enough to be left as is, but it's a good habit to define a named tuple type for the points used, as it partially deals with possible coordinate swapping errors. `type Point = tuple[x, y: int]` will be painless, you can simply replace the anonymous tuples in all the declarations without requiring any other changes in the code. Writing `foo.x` is always more pleasant than `foo[0]`.

This is all optional, the code looks good already!

ZoomRmc (approved):
LGTM!

dlesnoff (approved):
Nice!
I learnt a new algorithm :)

PR #73, opened 2024-05-04
chore: use nim version `2.0.8`

vil02:
[Nim 2.0.4](https://nim-lang.org/blog/2024/04/16/versions-1620-204-released.html) and [Nim 2.0.8](https://nim-lang.org/blog/2024/07/03/version-208-released.html) was relleased. This PR makes a suitable update.

Similar to #67.

### DIFF
--- a/.github/workflows/nim_test.yml
+++ b/.github/workflows/nim_test.yml
@@ -20,7 +20,7 @@ jobs:
           - macos-latest
           - windows-latest
         nim-version:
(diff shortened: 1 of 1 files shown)


dlesnoff:
Sorry, I was very busy to not notice this very simple pull request.
Could you please change again the last digit for 2.0.8 ?

dlesnoff (approved):

PR #74, opened 2024-05-05
chore: configure nimalyzer

vil02:
In this PR I propose to use [nimalyzer](https://www.laeran.pl/repositories/nimalyzer/wiki?name=About) as a static analysis tool for this repository. I add a workflow `nimalyzer.yml` and a config file `.nimalyzer.cfg`. Please notice that some of the _rules_ are temporarily commented out (with having them in, the check would simply fail). I suggest to create a issue, where we can discuss which rules to _activate_ and trace the progresses of fixing the warnings.

The config file is a simplified version of the [official example](https://github.com/thindil/nimalyzer/blob/7ee091ab1b69150c03f229eeb28f8da162aad295/config/nimalyzer.cfg). All of the `nim`-files are the object of the analysis.

Please notice that `nimalyzer.yml` will run one per week. Due to the fact that the tool is still under development, I think it would be a good idea to know early enough if something breaks or changes.

Related to #16.

### DIFF
--- a/.nimalyzer.cfg
+++ b/.nimalyzer.cfg
@@ -1,0 +1,69 @@
+files */*.nim
+showSummary true
+
+# TODO
+# check hasPragma all contractual "raises: [*"
+# explanation Contracts helps in testing the program and all declared procedures should have declared contracts for them. The procedures should avoid raising exceptions and handle each possible exception by themselves for greater stability of the program.
+
+check paramsUsed procedures
+explanation Unused parameters only clutter the source code and can cause confusion.
+
+check paramsUsed macros
+explanation Unused parameters only clutter the source code and can cause confusion.
+
+# TODO
+# check namedParams
+# explanation Named parameters allow avoiding assigning invalid values to the calls but also allow to assing the calls' parameters in arbitrary order.
+
+# TODO
+# check hasDoc all
+# explanation The documentation is a love's letter to your future self. :) Documentation make our lives easier, especially if we have return to the code after a longer period of time.
+
+# TODO
+# check varDeclared full
+# explanation The full declaration of variables gives information about their types and sets the initial values for them which can prevent sometimes in hard to detect errors, when the default values change.
+
+# TODO
+# check varUplevel
+# explanation The proper usage of var, let and const types of declaration make the code more readable and prevent from invalid assigning to a variable which shouldn't be assigned.
+
+check localHides
+explanation If a local variable has the same name as a global one declared in the same scope, it can lead to hard to read code or even invalid assign to the variable.
+
+# TODO
+# check ifStatements all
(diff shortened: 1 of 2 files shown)


dlesnoff:
I am sorry for my very late reaction.

It seems too early for me to integrate nimalyzer.
I think it would be better to wait a little bit before incorporating it.
I'll check locally the output of nimalyzer.

.nimalyzer.cfg
```diff
@@ -0,0 +1,69 @@
+files */*.nim
+showSummary true
+
+# TODO
+# check hasPragma all contractual "raises: [*"
+# explanation Contracts helps in testing the program and all declared procedures should have declared contracts for them. The procedures should avoid raising exceptions and handle each possible exception by themselves for greater stability of the program.
+
+check paramsUsed procedures
```
dlesnoff, line 8: On my side, I've got "No rule named 'paramsused' available".

nimble list -i | grep nimalyzer
nimalyzer  [(version: 0.9, checksum: 6719a737ef6041b25b6403f0a353f1846a891890)]

dlesnoff, line 8:
```suggestion
check params procedures
```

.nimalyzer.cfg
```diff
@@ -0,0 +1,69 @@
...
+# check assignments shorthand
+# explanation Shorthand assignments are shorter to write and can be more readable, especially with long names of variables.
+
+check caseStatements min 3
+explanation Short case statements can be replaced by if statements for better readablity.
+
+check ifStatements max 3
+explanation Long if statements can be replaced by case statements for better readability.
+
+check complexity cyclomatic all 40
```
dlesnoff, line 58:
```suggestion
check complexity all 40
```
0.9v change

vil02:
1f7a2d0c5c3c17c1e4e048bb27d1931c3e5f4d6e is similar to #78.

Regarding the _early stage_ of nimalyzer: are there some alternative tools? I would use it as long as it would not enforce some strange workarounds. We could also _pin_ the version by installing it with a command like
```yaml
nimble install nimalyzer@0.9 -y
```

dlesnoff:
The output of nimanalyzer is strange:
"[11:37:32] ERROR  rule: params, procedure constructHighway* line: 27 parameter 'randomFrequency' doesn't use int or string as type."
I don't understand the "doesn't use int or string as type.". I understand from it that the nimanalyzer tool doesn't support finding unused parameters if there are not of these types.

I don't know any alternatives for this kind of tool.

dlesnoff:
Re-read the documentation of Nim analyzer, there is a bug with the rule:
"check params used all".

Doc: https://www.laeran.pl.eu.org/repositories/nimalyzer/wiki?name=paramsUsed%20rule
Since both checkType and declarationType may use the keyword "all", there might be a parser error here, that replaces "used" by "all" for the checkType (when the "all" is actually a declarationType thing). I'll open an issue in the Fossil repository if I manage that.

PR #82, opened 2024-10-04
Finds the mean of a list of numbers

Bigbublik:


ZoomRmc:
Please, read this repo's [contribution guidelines](/TheAlgorithms/Nim/blob/main/CONTRIBUTING.md). If you had followed them, you'd know that your code doesn't work in its current form - the tests would show it.

1. The file should include a module-level doc-comments, the minimal requirement is the module's name in the first line for auto populating the [directory](/TheAlgorithms/Nim/blob/main/DIRECTORY.md).
2. Examples used in the documentation should be [runnableExamples](https://nim-lang.org/docs/system.html#runnableExamples%2Cstring%2Cuntyped) for additional testing. One example is enough, the rest should be tests.
3. The module lacks tests. Follow the [minimal example](/TheAlgorithms/Nim/blob/main/CONTRIBUTING.md#minimal-example) and merged algorithms on how to implement them.
4. `nums` should be an `openArray[SomeNumber]`. You'll have to use a loop or `std/sequtils.foldl`, or import the `std/math.sum` which works on openArrays.

dlesnoff:
Sums had a specific module in the stdlib, but were changed to a nimble package.

Anyway, I would keep a similar version with `openArray[SomeFloat]` and maybe have a more complex version with `openArray[SomeNumber]`. I might only keep the second version but I would like to compare both.

I second @ZoomRmc : tests from the documentation could be changed copied under the when isMainModule.
Please, mark your PR as draft if you are not done. The following code indicates so:
```nim
when isMainModule:
  discard
```

maths/average_mean.nim
```diff
@@ -0,0 +1,30 @@
...
+  ## # Output: 20.0
+  ## ```
+  ## ```nim
+  ## echo mean(@[1, 2, 3, 4, 5, 6, 7, 8])
+  ## # Output: 4.5
+  ## ```
+  ## ```nim
+  ## echo mean(@[])
+  ## # Error: List is empty
+  ## ```
```
dlesnoff, line 24: Move this under the when isMainModule block.

dlesnoff (changes requested):
See my answer in the conversation tab.

PR #84, opened 2024-10-08
Fix and clean CI after Nim 2.2.0 update

dlesnoff:
- **Specify template's return type for correct variable declaration**
I specified the return type of a template in strings/check_anagram.nim
There was an error about non-discarded char elements, as somehow these values were not considered part of the declaration.
I am not super fan of this fix, as one branch of the case switch just errors out and do not return a value. I am all ears for a better fix.

- **[modular_inverse] Reduce the number of tests to make CI readable again**
The CI was not very readable due to a too great number of tests in the maths/modular_inverse.nim file. I reduced the number of tests from 1000 to 10.

### DIFF
--- a/maths/modular_inverse.nim
+++ b/maths/modular_inverse.nim
@@ -67,7 +67,7 @@ when isMainModule:
       check modularInverse(17, 17).is_none()
 
     randomize()
-    const randomTestSize = 1000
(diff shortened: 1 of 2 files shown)


vil02 (approved):

dlesnoff:
Thanks!

